PR #441, opened 2026-05-15
fix: strip joinorder primary keys for presto

joeharris76:


joeharris76:
Superseded by #446, which rebases the Presto DDL fix onto current develop, keeps the fast-lane test count under policy, and includes the Spark JoinOrder config-path follow-up from review.

PR #478, opened 2026-05-20
test: enforce import dependency contract

joeharris76:
## Summary
- replace the pandas-only dependency assertion with a runtime-loaded import contract
- inspect top-level imports from eager BenchBox modules for `import benchbox` and `benchbox --help`
- assert direct eager third-party imports map to core dependencies and optional-only deps stay off eager paths

## Verification
- `uv run -- python -m pytest tests/unit/test_package_dependencies.py -q`
- `uv run --python 3.10 -- python -m pytest tests/unit/test_package_dependencies.py -m fast -q`
- `uv run --python 3.12 -- python -m pytest tests/unit/test_package_dependencies.py -m fast -q`
- temporarily removed the core `pandas>=2.0.0` line; tests failed on `pandas` as undeclared/optional-only, then `pyproject.toml` was restored
- `uv run --project ~/.claude/tools/todo todo-index`
- `uv run --project ~/.claude/tools/todo todo-cli check-graph`

## Notes
- This intentionally does not build or install wheels; wheel isolation is covered by PR #476.
- `todo-validate --all` is unavailable in this checkout because it resolves a missing shared schema path.

PR #479, opened 2026-05-20
ci: add hermetic wheel smoke gates

joeharris76:
## Summary
- replace the pre-merge package smoke with an off-project `uv run --isolated --no-project --with <wheel>` import and CLI check
- add a release pre-publish smoke job between build and publish
- pin post-publish installation smoke to `benchbox==${{ needs.build.outputs.version }}`

## Verification
- `uv run -- python -c "import yaml; [yaml.safe_load(open(f)) for f in ['.github/workflows/test.yml','.github/workflows/release.yml']]; print('yaml ok')"`
- `uv build --wheel --out-dir /tmp/bbw && cd "$(mktemp -d)" && uv run --isolated --no-project --with /tmp/bbw/*.whl -- python -c "import benchbox; assert 'site-packages' in benchbox.__file__, benchbox.__file__; print(benchbox.__version__)"`
- `cd "$(mktemp -d)" && uv run --isolated --no-project --with /tmp/bbw/*.whl -- benchbox --help`
- pandas-stripped temp wheel fails with `ModuleNotFoundError: No module named 'pandas'` via `benchbox/platforms/clickhouse/client.py:11`
- `uv run --project ~/.claude/tools/todo todo-index`
- `uv run --project ~/.claude/tools/todo todo-cli check-graph`

## Notes
- Replaces closed PR #476; this head branch uses the release-style branch name required by the main-base validator.
- `todo-validate --all` is unavailable in this checkout because it resolves a missing shared schema path.

PR #481, opened 2026-05-20
test: enforce import dependency contract

joeharris76:
## Summary
- replace the pandas-only dependency assertion with a runtime-loaded import contract
- inspect top-level imports from eager BenchBox modules for `import benchbox` and `benchbox --help`
- assert direct eager third-party imports map to core dependencies and optional-only deps stay off eager paths

## Verification
- `uv run -- python -m pytest tests/unit/test_package_dependencies.py -q`
- `uv run --python 3.10 -- python -m pytest tests/unit/test_package_dependencies.py -m fast -q`
- `uv run --python 3.12 -- python -m pytest tests/unit/test_package_dependencies.py -m fast -q`
- temporarily removed the core `pandas>=2.0.0` line; tests failed on `pandas` as undeclared/optional-only, then `pyproject.toml` was restored
- `uv run --project ~/.claude/tools/todo todo-index`
- `uv run --project ~/.claude/tools/todo todo-cli check-graph`

## Notes
- Replaces closed PR #478; this head branch uses the release-style branch name required by the main-base validator.
- This intentionally does not build or install wheels; wheel isolation is covered by PR #479.
- `todo-validate --all` is unavailable in this checkout because it resolves a missing shared schema path.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -53,13 +53,8 @@ jobs:
       - name: Compatibility docs drift check
         run: make compat-docs-check
 
-      - name: Dependency audit (no unused declarations)
-        run: |
-          if [ -d _project/scripts ]; then
-            make audit-deps
-          else
-            echo "_project/scripts/ is absent on curated release branches; skipping dependency audit."
-          fi
+      - name: Dependency audit (declared/imported contract)
+        run: make audit-deps
 
       - name: Release curation list drift check
         run: |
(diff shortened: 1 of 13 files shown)


joeharris76:
Batch note: implementation and local verification are complete, and the formatting CI failure was fixed in commit 112d8caf7. `perf-smoke` failed twice, including the one allowed retry, despite the comparison reporting overall SIGNIFICANT_IMPROVEMENT / ~30% faster total runtime. The failures are two per-query micro-regressions over the 10% threshold, which is outside the import-contract TODO scope. Leaving this PR open but blocked on the perf-smoke gate.

PR #709, opened 2026-05-30
fix/drop pandas core deps

joeharris76:
- **fix(deps): remove pandas from the import benchbox surface**
- **fix(deps): drop pandas from core deps and gate the minimal import surface**
- **refactor(deps): align clickhouse lazy __init__ with house pattern; trim CI gate**


joeharris76:
@codex review

PR #756, opened 2026-06-06
fix(tpchavoc): qualify correlated-subquery self-binding in variants

joeharris76:
## Summary

Implements the `tpchavoc-correlated-subquery-self-binding` TODO.

Three TPC-Havoc variant queries carried a correlated subquery whose **unqualified** correlation column silently bound to the **inner** alias instead of the outer table, degenerating the intended per-row correlation into an uncorrelated scan — the same defect class fixed in `10_v8` (uat-duckdb-reference-triage). Because TPC-Havoc validates execution rather than result equivalence, the defect was silent.

The fix qualifies each correlation to the outer (unaliased) table, matching the `10_v8` approach, **without altering the variant's SQL shape** (per the TODO's anti-pattern guidance).

## Fixes

| Variant | Subquery | Before | After |
|---|---|---|---|
| `q11_v1` | scalar threshold | `ps2.ps_partkey = ps_partkey` | `ps2.ps_partkey = partsupp.ps_partkey` |
| `q17_v8` | `EXISTS` | `avg_calc.l_partkey = l_partkey` | `avg_calc.l_partkey = lineitem.l_partkey` |
| `q2_v8` | `NOT EXISTS` (min-cost) | `ps2.ps_supplycost < ps_supplycost` | `ps2.ps_supplycost < partsupp.ps_supplycost` |

`q11_v1` and `q17_v8` are the two cases named in the TODO. The required **sweep of all variant sets** (work item `w3`) turned up a third: `q2_v8`, a `NOT EXISTS` rewrite of Q2's "minimum cost supplier" whose `ps_supplycost` self-bound to `ps2`, making the predicate always false and the `NOT EXISTS` always true.

In `q17_v8` the sibling `l_quantity` reference is also qualified to `lineitem.l_quantity`: it is currently bound correctly only because `avg_calc` happens not to expose `l_quantity`, so qualifying it hardens the variant against the exact same self-binding class.

## Verification

Against real TPC-H SF=0.01 data (DuckDB):

- **`q2_v8`**: buggy form returned **5 rows**; fixed form returns **4 rows**, matching the canonical `q2_v1`/`q2_v3` — i.e. this variant was returning semantically wrong results, now corrected.
- **`q11_v1`** / **`q17_v8`**: fixed forms match their canonical siblings; at small scale the degenerate form coincidentally produced the same result (the "silent" defect the TODO describes). Correctness lives in the correlation binding, covered by the new assertions.
- Full DuckDB `tpchavoc` run (`load,power`, SF 0.01): **220/220** queries execute, no failures.
- `pytest tests/integration/test_tpchavoc_duckdb.py` and the `tpchavoc` unit suites pass; `ruff check`/`format` clean.

## Tests

`tests/integration/test_tpchavoc_duckdb.py`:
- The three variants join the regression `EXPLAIN` parse/bind list.
- New `test_correlated_subquery_variants_bind_to_outer_table` asserts each correlation is qualified to the outer table and that `q11_v1` decorrelates into a DuckDB delimiter (`DELIM`) join. `q17_v8`/`q2_v8` are guarded by the substring assertion only — each retains a *second* qualified correlation, so their plans look decorrelated even in the buggy form and cannot distinguish fixed from broken (documented inline).
- Extracted the repeated DuckDB schema-setup into a shared `_create_duckdb_schema` helper.

Changes are confined to the TODO's `scope_limit.only_modify` paths: `benchbox/core/tpchavoc/variant_sets/` and `tests/integration/test_tpchavoc_duckdb.py`.

https://claude.ai/code/session_01Mpw7mc1gcoQVTtz4HQZwyk

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mpw7mc1gcoQVTtz4HQZwyk)_

### DIFF
--- a/benchbox/core/tpchavoc/variant_sets/q02.yaml
+++ b/benchbox/core/tpchavoc/variant_sets/q02.yaml
@@ -397,7 +397,7 @@ variants:
               and s2.s_nationkey = n2.n_nationkey
               and n2.n_regionkey = r2.r_regionkey
               and r2.r_name = 'EUROPE'
-              and ps2.ps_supplycost < ps_supplycost
+              and ps2.ps_supplycost < partsupp.ps_supplycost
         )
     order by
         s_acctbal desc,
--- a/benchbox/core/tpchavoc/variant_sets/q11.yaml
+++ b/benchbox/core/tpchavoc/variant_sets/q11.yaml
@@ -20,7 +20,7 @@ variants:
             where ps2.ps_suppkey = s2.s_suppkey
               and s2.s_nationkey = n2.n_nationkey
               and n2.n_name = 'GERMANY'
-              and ps2.ps_partkey = ps_partkey
+              and ps2.ps_partkey = partsupp.ps_partkey
         )
     group by
         ps_partkey
--- a/benchbox/core/tpchavoc/variant_sets/q17.yaml
+++ b/benchbox/core/tpchavoc/variant_sets/q17.yaml
@@ -30,7 +30,7 @@ variants:
 - id: 8
   description: 'EXISTS pattern: Use EXISTS for membership testing'
   sql: |
-    select sum(l_extendedprice) / 7.0 as avg_yearly from lineitem, part where p_partkey = l_partkey and p_brand = 'Brand#23' and p_container = 'MED BOX' and exists (select 1 from (select l_partkey, 0.2 * avg(l_quantity) as threshold from lineitem group by l_partkey) avg_calc where avg_calc.l_partkey = l_partkey and l_quantity < avg_calc.threshold)
+    select sum(l_extendedprice) / 7.0 as avg_yearly from lineitem, part where p_partkey = l_partkey and p_brand = 'Brand#23' and p_container = 'MED BOX' and exists (select 1 from (select l_partkey, 0.2 * avg(l_quantity) as threshold from lineitem group by l_partkey) avg_calc where avg_calc.l_partkey = lineitem.l_partkey and lineitem.l_quantity < avg_calc.threshold)
 - id: 9
   description: 'Window functions (OLAP): Use window functions for ranking'
   sql: |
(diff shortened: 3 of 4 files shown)


joeharris76:
## Full-scale verification + semantic-equivalence evidence

Beyond the SF0.01 smoke in the description, I verified at **SF1.0** that these changes serve the benchmark's purpose (structurally diverse but semantically **equivalent** rewrites of each TPC-H query) rather than working against it.

### Execution (`must_preserve`: 100% green)
- SF0.01: `load,power` → **220/220** execute, no failures.
- SF1.0: `power` → **220 queries × 4 runs = 880/880, 100% success, PASSED.**

### Equivalence vs the canonical TPC-H query (the real invariant, checked against the base query — not sibling variants)
At SF1.0, against the populated DuckDB:

| Variant | Result |
|---|---|
| `q11_v1` | **exact set-match** to canonical TPC-H Q11 (1048 rows) |
| `q17_v8` | **exact set-match** to canonical TPC-H Q17 |
| `q2_v8` | **exact set-match** to canonical TPC-H Q2 *without its `LIMIT 100`* (460 rows) |

The `LIMIT 100` is omitted by the **entire q2 variant family** (pre-existing design, untouched here). Critically, all ten q2 variants now return 460 rows — the fix brought `q2_v8` **into** its sibling equivalence class; pre-fix it was the odd one out (always-true `NOT EXISTS` → extra rows).

### Shape faithfulness
EXPLAIN confirms the variants are now *genuinely correlated*, not degenerate: `q11_v1` decorrelates into a DuckDB delimiter join (pre-fix it folded to an uncorrelated aggregate). A variant labelled "Scalar subqueries"/"EXISTS"/"NOT EXISTS" now actually presents that correlated shape to the optimizer — which is what TPC-Havoc exists to measure. The pre-fix forms were mislabelled (uncorrelated) tests.

### Unrelated observation (out of scope for this PR)
While cross-checking, `q2_v10` (CASE-expressions variant) returns a *different* result set from the other nine q2 variants — a pre-existing equivalence discrepancy unrelated to the self-binding pattern and not touched here. Likely its `min(case when ps_supplycost > 0 then … else 999999 end)` reformulation isn't faithful. Flagging for a separate TODO.

https://claude.ai/code/session_01Mpw7mc1gcoQVTtz4HQZwyk

---

PR #812, opened 2026-06-16
feat(audit): durable self-binding detector + cross-surface correlated-subquery audit

joeharris76:
## Summary

Implements `correlated-subquery-self-binding-cross-surface-audit`. PR #756 hand-fixed three TPC-Havoc SQL variants where an **unqualified correlation column silently bound to the inner relation**, degenerating a per-row correlation into an uncorrelated scan (wrong results, execution stays green). That sweep was manual and SQL-only. This adds a **durable detector** so the class can't silently reappear, and audits the surfaces the manual sweep never covered.

## What's here

- **`_project/scripts/detect_self_binding.py`** — a sqlglot-backed, **catalog-aware** advisory detector. It builds its column→table index from the benchmark's own `CREATE TABLE` DDL (so it's benchmark-agnostic), and flags a bare column in a subquery comparison **only when** (a) the other operand is a column qualified by an inner source alias (the correlation shape), **and** (b) the bare column is provided by both an inner and an outer source (the ambiguous bind that makes the defect silent). It recognizes derived-table/CTE output columns, so CTE correlation targets — the #756 Q17 case — are covered.
- **`tests/unit/test_self_binding_detector.py`** — the durable guard: flags all three #756 cases when reverted, stays clean on the fixed forms and all 220 shipped TPC-Havoc variants, and ignores plain filters and fully-qualified correlations.

## Audit results (detector swept at SF=1, DuckDB dialect)

| Surface | Queries | Candidates |
|---|---|---|
| TPC-Havoc SQL | 220 | 0 (already fixed by #756) |
| SSB | 13 | 0 |
| join-order | 113 | 0 |
| read-primitives | 153 | 0 |
| write-primitives | 109 | 0 |
| TPC-DS | 99 | 2 → **false positives** |

The two TPC-DS flags (q14, q54) are human-confirmed **false positives**: derived-table aliases that reuse a base-table column name (`item.i_item_sk AS ss_item_sk`; `ws_bill_customer_sk AS customer_sk`), so the bare column binds correctly. These are official TPC-DS spec queries and were **not modified** (conformance).

**TPC-Havoc DataFrame variants** are covered by the merged DataFrame equivalence gate (#785), which is green — an inner-frame self-bind would surface there as a divergence. No drift found.

**Conclusion:** no real self-binds exist outside the already-fixed #756 cases. The detector is **advisory** (not a hard CI blocker), per the TODO's anti-pattern guidance ("don't ship a noisy static check as a hard gate").

## Verification
- `python -m pytest tests/unit/test_self_binding_detector.py` — 6 passed.
- `ruff check` / `ruff format --check` / `ty check` — clean.
- TODO ledger moved to `active/`, w0–w3 marked done with findings, status Under Review.

## Scope notes
- The detector lives in `_project/scripts/` (an audit/lint tool, not shipped in the package), within the TODO's `scope_limit`.
- Provenance open-question: the detector found **no** new instances across six benchmarks, which argues the #756 cases were an isolated hand-authored set rather than a systematic generator artifact — so a generator correctness-contract (the TODO's deferred item) is not warranted on this evidence.

https://claude.ai/code/session_01Mpw7mc1gcoQVTtz4HQZwyk

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mpw7mc1gcoQVTtz4HQZwyk)_

### DIFF
--- a/_project/TODO/main/active/correlated-subquery-self-binding-cross-surface-audit.yaml
+++ b/_project/TODO/main/active/correlated-subquery-self-binding-cross-surface-audit.yaml
@@ -2,7 +2,7 @@ id: correlated-subquery-self-binding-cross-surface-audit
 title: "Audit DataFrame variants and other benchmarks for correlated-subquery self-binding, and add a durable detector"
 worktree: main
 priority: Medium
-status: Not Started
+status: Under Review
 category: Testing and Quality Assurance
 
 description: |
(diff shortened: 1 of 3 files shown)


joeharris76:
Good catch — fixed in `47a06611`. Confirmed the gap: `build_scope` doesn't expose an `UPDATE`/`DELETE` target as a parent scope of its subqueries, so the `parent is None` skip left correlated DML subqueries unanalyzed — the write-primitives "clean" result was a blind spot, not a verified pass.

Fix: `_dml_outer_columns` now treats the DML target (and `USING`/`FROM`) tables as the outer correlation source, and I dropped the now-redundant parentless-scope skip (the "providable by an outer source" requirement already prevents top-level false positives). A correlated self-bind inside `UPDATE`/`DELETE` is now flagged (added regression tests for both); the qualified-to-target form stays clean; and the full sweep is unchanged — write-/read-primitives are still 0, but now genuinely analyzed rather than skipped.

---

PR #851, opened 2026-06-21
docs(equivalence): file + verify ClickBench cross-surface follow-up TODO

joeharris76:
## Summary

Running the cross-surface SQL↔DataFrame equivalence gate (cross-surface campaign, w2/w3) over **ClickBench** in report mode (SF=0.1, 86 cells → ~35 divergent) surfaced a finding that reframes whether ClickBench can be exact-gated at all. This PR captures the rigorously re-evaluated findings as a planning TODO, plus the load-faithful report scaffolding needed to reproduce them. **No CI gate is wired** (no test or Make target invokes the new builder), so it stays non-blocking.

### Root-cause breakdown of the 35 divergent cells
- **(A) Real loader bug — empty string → null/NaN on `.tbl`/`.csv` load (high value, all benchmarks).** DuckDB keeps `''` in VARCHAR columns; both DataFrame loaders read empty fields as null/NaN. Verified directly: `SearchPhrase` `''`=74948 / NULL=0 in DuckDB vs null/NaN=74948 in both backends. Reliable symptom: `Q6_pandas` `COUNT(DISTINCT)` off by one.
- **(B) Real impl bug — expr date column compared to a string literal** in the shared july mask (`queries.py:31`); the pandas twin already parses via `date.fromisoformat`. Q37/Q38/Q43.
- **(C) Real impl bug — missing unified-wrapper methods**: `UnifiedLazyFrame.slice` (OFFSET path, Q39–42) and `UnifiedStrExpr.replace` (Q29).
- **(D) Not a bug — irreducible `ORDER BY`/`LIMIT` tie non-determinism** (dominant, ~15 queries): values match, but each engine returns a different *valid* tie order / row-set, so exact positional comparison can't pass them.
- **(E) Not a bug, gateability blocker — the ClickBench generator is unseeded**, so each build yields different data and a different exact divergent-cell set.

A dtype check **disproved** an earlier "big-int loaded as float64 → precision loss" hypothesis: UserID/WatchID/ClientIP load faithfully with identical `n_unique` across all three surfaces.

**Consequence/recommendation (for w3 to ratify):** land the loader fix standalone (benefits ssb/coffeeshop too), fix the two expr impl bugs, and stand up the clean exact gate on **joinorder_synthetic** rather than force ClickBench green — classify-or-descope its tie/LIMIT queries.

## Review pass (this PR includes the corrections)
An independent code review + a re-build corrected three things from the first draft:
- `read_csv` refs pointed at abstract signatures → fixed to the concrete sites (`pandas_df.py:316`, `polars_df.py:291`).
- Q25/Q27_pandas reframed from "reliable" to **data-dependent** (a re-build ran them clean).
- Added finding (E) (unseeded generator); w0/w1/verification now assert structure over the varying exact cell list.

## Changes
- `_project/TODO/main/planning/clickbench-cross-surface-tie-nondeterminism-and-loader.yaml` — new TODO (schema-validated).
- `benchbox/core/equivalence/cross_surface.py` — adds `build_clickbench_duckdb` (mirrors `build_ssb_duckdb`) + registers `clickbench` in `GATES` for report mode only.

## Verification
- TODO passes `_project/scripts/validate_todo.py`.
- Empty-string→null, dtype parity, and the per-query SQL/categorization claims verified against a live build; file/line refs confirmed by an independent review agent.
- No test iterates `GATES`; import + key-set checked, no CI gate added.

---
_Generated by [Claude Code](https://claude.ai/code/session_017nZBWGFNZeZGHDm3usxyev)_

### DIFF
--- a/_project/TODO/main/planning/clickbench-cross-surface-tie-nondeterminism-and-loader.yaml
+++ b/_project/TODO/main/planning/clickbench-cross-surface-tie-nondeterminism-and-loader.yaml
@@ -1,0 +1,259 @@
+id: clickbench-cross-surface-tie-nondeterminism-and-loader
+title: "ClickBench cross-surface gate: empty-string loader fidelity, expr impl bugs, and irreducible ORDER BY/LIMIT tie non-determinism"
+worktree: main
+priority: High
+status: Identified
+category: Testing and Quality Assurance
+
+description: |
+  Follow-up from the cross-surface equivalence gate campaign
+  (benchmark-cross-surface-equivalence-gate, w2/w3). ClickBench is one of the
+  two benchmarks the applicability sweep marked GATEABLE (clickbench 43/43,
+  joinorder_synthetic 13/13). Running the cross-surface gate in REPORT mode over
+  ClickBench at SF=0.1 on the production-loader substrate produced
+  35 divergent cells across 86 (43 queries x 2 backends: polars expression +
+  pandas). A careful re-evaluation (running each divergent query's SQL reference
+  vs each DataFrame impl on the SAME generated data, inspecting full result rows
+  and column dtypes) splits those 35 cells into FOUR root causes - three are
+  genuine, fixable bugs; the fourth is irreducible and reframes whether ClickBench
+  can be exact-gated at all.
+
+  EVIDENCE (report run, SF=0.1, 86 cells, 35 divergent):
+    GATE FAILURE - 23 distinct queries:
+      Q6, Q9, Q10, Q15, Q16, Q17, Q18, Q19, Q24, Q25, Q27, Q29, Q31, Q32, Q33,
+      Q36, Q37, Q38, Q39, Q40, Q41, Q42, Q43.
+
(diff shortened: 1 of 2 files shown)


joeharris76:
Closing without merging — superseded by `develop`.

`develop` already carries the load-faithful `build_clickbench_duckdb` builder, correctly registered in `STAGED_GATES` (not `GATES`). This PR was branched from an older `develop` and would instead:

- re-add `build_clickbench_duckdb` (now a merge conflict — `mergeable_state: dirty`), and
- promote `clickbench` from `STAGED_GATES` into `GATES`, which breaks `test_clickbench_is_staged_not_enforced` and contradicts the cross-surface remediation TODO's own anti-pattern ("do not promote a staged gate to GATES while it is nondeterministic").

The substantive findings (empty-string→null loader bug, expr/date and unified-wrapper impl bugs, tie non-determinism, unseeded generator) are already tracked in `cross-surface-oracle-remediation.yaml` (w1, w3, w5, w6) and will be addressed there. Any unique detail from this PR's planning doc will be folded into w5's analysis update. No code lost.

---

PR #883, opened 2026-06-24
fix(equivalence): harden the order-aware boundary-tie comparator (>=2 guard + LIMIT N+1 probe)

joeharris76:
## Summary

Resolves the order-aware path of `cross-surface-comparator-boundary-tie-soundness` (#880) in two steps — a correctness fix, then the precise probe that supersedes the heuristic it relied on.

### 1. `>= 2` guard on the order-aware path (the masking bug)
A self-review found `ResultValidator._validate_order_aware` accepted **any** final-group membership difference under a trailing `LIMIT` (`if i == last_index and tie_aware: continue`), with **no minimum-tie guard** — so a deterministic *single-row* final group with a wrong value was silently masked. The order-*blind* heuristic already had a `>= 2` guard (#878), and the #880 TODO even cited `_validate_order_aware` as the guard site, but the guard was never there. Fixed by requiring `len(orig_rows) >= 2`.

### 2. `LIMIT N+1` boundary-tie probe (the precise signal — #880 w1)
The `>= 2` heuristic is imprecise in **both** directions:
- a **one-visible-row** truncated tie → false positive (flagged, shouldn't be);
- a **≥2-visible *complete* (untruncated)** final group that differs → false negative (masked real bug).

`cross_surface._boundary_tied_past_limit` re-runs each trailing-`LIMIT` query's SQL reference at `LIMIT N+1` and checks whether the first dropped row shares the final kept row's resolved order-key value:

| Probe result | Meaning | Final-group rule |
|---|---|---|
| `True` | order key tied past the cutoff | accept the swap (even one visible row) |
| `False` | final group is complete/untruncated | must match exactly |
| `None` | not applicable / unanswerable | fall back to the `>= 2` heuristic |

`validate_results_exact` / `_validate_order_aware` gain `boundary_tie_past_limit`; the gate threads the probe per query.

## Verification
- 100 validation + cross-surface unit tests pass, incl. new coverage for the comparator signal (`test_boundary_probe_*`) and the probe against real DuckDB (`TestBoundaryTiedPastLimitProbe`) — covering the `keys 10,5,5 LIMIT 2` boundary-tie (now tolerated) and `(2,'good')/(2,'bad')` deterministic case (caught), per the TODO.
- All five enforced gates stay green: clickbench / ssb / amplab / joinorder_synthetic confirmed "equivalent"; coffeeshop is divergence-free, so the differ-only path is inert.
- **Additive**: `None` preserves prior behavior; TPC-Havoc (never sets `order_aware`) stays byte-identical (440 cells, 0 divergent).

## Scope
`benchbox/core/tpchavoc/validation.py` (guard + `boundary_tie_past_limit`), `benchbox/core/equivalence/cross_surface.py` (the probe + wiring), tests, and the #880 TODO (w1 marked done for the order-aware path; the order-blind-heuristic residual — unmappable `ORDER BY`/`SELECT *` only, latent — is recorded as deferred).

Context: this carries forward the one genuinely-unaddressed finding from my review of the (already-landed) cross-surface remediation; the rest were fixed on `develop` or tracked.


joeharris76:
Closing as superseded by #886. Both PRs implement the same #880 w1 boundary-tie work, and #886 landed on `develop` first with the identical two-part fix:

- the `>= 2` minimum-tie guard on the order-aware final group, **and**
- the `LIMIT N+1` boundary-tie probe — develop has `ResultValidator.validate_results_exact(..., final_key_tied_beyond_limit=)` with `validation.py` accepting the final group iff `len(orig_rows) >= 2 or final_key_tied_beyond_limit`, fed by `cross_surface._final_key_tied_beyond_limit` (the one-row-wider re-run).

#883's version (`boundary_tie_past_limit` / `_boundary_tied_past_limit`) is a competing implementation of the now-landed behaviour, so merging it would conflict with / partly revert #886. The residual order-blind-heuristic gap #883 noted is already tracked on `develop` as #890 ("residual >=2 final-tie-group soundness gap, follow-up to #886"). Nothing here is lost — closing rather than reworking the duplicate.

---

PR #947, opened 2026-07-03
feat(dev): opt-in `make ci-linux` Apple container Linux CI-parity sandbox

joeharris76:
## What

Opt-in `make ci-linux` that reproduces the Linux `pr.yml` gate inside an Apple
`container` machine on Apple silicon, plus the `docker/benchbox-agent` image it
runs in and AGENTS.md docs. Implements TODO `adopt-apple-container-ci-parity`.

## Why — w4 decision gate = **DIVERGES** (central finding)

Measured on **identical DuckDB 1.3.2 / arm64**, TPC-H SF=1, reference seed:

- **Linux** (container machine): `make test-correctness-gate` **PASS** — 18/18 value digests match the pinned reference.
- **macOS** native: **FAIL** on Q2/Q10/Q15 (same row count, different value digests; deterministic across 2 runs).
- `ssb` cross-surface equivalence: **PASS on both** (self-referential SQL↔DataFrame → OS-insensitive).

| query | Linux == pinned ref | macOS actual |
|---|---|---|
| Q2  | `74defe99…` | `2fafe9de…` |
| Q10 | `5489075d…` | `b2b0b469…` |
| Q15 | `dc79db98…` | `3e6e171e…` |

The pinned correctness-gate references are effectively **Linux-specific**, so a Mac
dev can't validate that soundness gate locally — this sandbox is the only way.

## Evidence gates (compact)

- **w0 PASS**: `container` 1.0.0, apiserver running, arm64 / macOS 26.5, mocker 0.5.4.
- **w1/w2**: machine boots only with an init in the rootfs (`systemd-sysv` → `/sbin/init`; minimal ubuntu fails); home-mount exposes the worktree **rw at its Mac path** via virtiofs.
- **w3**: correctness-gate + ssb equivalence ran inside the machine.

## Scope / caveats

- Purely additive & opt-in: no-op off Apple-silicon/macOS-26, **not** wired into any CI or `pr-open` target; `.github/workflows/` and `test-docker-*` untouched.
- Full `uv sync --group dev` + `make ci-local` inside the VM exceeded this host's free disk (the VM sparse disk only grows); w3/w4 used a targeted lock-pinned env (duckdb 1.3.2). The digest is DuckDB-build-tied, so this does not weaken the finding.
- Mocker for `test-docker-*` is tracked separately (`migrate-test-docker-stacks-to-mocker`).

## Review notes (deferred, by design)

Code review applied; these findings were intentionally not changed:
- **Base image / apt packages unpinned** — deliberate: it mirrors CI's own
  `ubuntu-latest`, which is likewise unpinned; pinning would drift *away* from the
  thing being reproduced.
- **`uname -m` no-ops under Rosetta** (an x86_64 shell on Apple silicon) — fail-safe:
  it skips with a message rather than misbehaving.
- **`sw_vers` non-numeric / single-quote in a `CI_LINUX_CMD` override** — theoretical /
  user-error only; shipped defaults and documented usage are unaffected.


joeharris76:
@codex review

PR #949, opened 2026-07-03
feat(test): opt-in CONTAINER_ENGINE=mocker for local test-docker-* stacks

joeharris76:
## What

Opt-in `CONTAINER_ENGINE` seam so `make test-docker-*` can run the existing
`docker/*/docker-compose.yml` stacks on Apple `container` via **Mocker** instead of
Docker Desktop, plus a compose-lifecycle parity harness. Implements TODO
`migrate-test-docker-stacks-to-mocker` (builds on the Apple `container` runtime from #947).

**Default stays `docker`; CI is unchanged.** `CONTAINER_ENGINE=mocker` is Apple-silicon /
macOS-26 local-dev only and must never run in CI.

## Why / the gate (w4 = option **b**, teardown-wrap)

w0 re-confirmed the spike: mocker 0.5.4's `compose down -v` removes containers but
**leaks named volumes** (a stale-data correctness risk across benchmark runs).
Since no upstream fix, w4 resolves via **(b)**: a Makefile teardown wrap
(`compose_down_fresh`) prunes the project's named volumes after `down -v` — a no-op on
docker (which already removes them), scoped to the project so it never touches
unrelated volumes.

## Evidence

- **w0**: mocker 0.5.4 / container 1.0.0; questdb `up --wait` health-gated + `localhost:9003`→200; postgres `down -v` **leaked** `bbpg-postgresql18-data` (leak persists).
- **w1 parity harness** (`_project/scripts/mocker_compose_parity.sh`): asserts health-gated up, published-port reachability, and no-container/no-volume after teardown.
- **Parity MEASURED** — `make test-docker-parity` PASS on **both** engines (questdb + postgresql): docker and mocker produce identical lifecycle outcomes; the wrap makes mocker's post-`down -v` state match docker's.
- **w3 end-to-end**: `CONTAINER_ENGINE=mocker make test-docker-questdb` — 3 `live_questdb` tests load parquet data + query through the connector against the mocker-hosted stack, **pass** (42.98s); clean teardown.
- **w2 multi-service networking**: databend (minio + minio-setup + databend, `depends_on` conditions) — mocker honors ordering (`up --wait` 4/4) but the stack **does not stay healthy** (minio exits, databend :8000 empty reply); the **same stack on docker stays healthy and serves 200**. So 3+ service stacks are not reliable on mocker → use docker for multi-service.

## Scope / caveats

- Only `Makefile`, `AGENTS.md`, `_project/scripts/mocker_compose_parity.sh` changed. `docker/*/docker-compose.yml` untouched (stay portable Compose specs for docker + CI). `.github/workflows/` untouched.
- Default not flipped: docker remains default; mocker is opt-in and reversible via the variable.
- doris/starrocks are all-in-one single-container images (2 named volumes each; the wrap covers volume cleanup); databend is the genuine cross-container stack (databend→minio).

## Review notes

Two independent reviews applied. Fixed: header/column-agnostic volume grep in both the wrap and harness (was a correlated `awk 'NR>1{print $NF}'` blind spot); generic `port_serves` now resolves `${VAR:-default}` port syntax (doris/starrocks); empty project name now FAILs instead of silently passing; interrupt trap tears down an in-flight stack; `tcp_ok` fd cleanup. Deferred (low/by-design): `pg-extensions` and other specialized live-test targets stay docker-only (out of the `DOCKER_PLATFORMS` scope); the postgresql `:5432` reachability probe is host-scoped (theoretical collision with a host Postgres); the generic fallback probes the first published port.


joeharris76:
@codex review

PR #1067, opened 2026-07-09
fix: normalize string option-spec defaults through their declared parser

joeharris76:
get_default_options() returned spec-row defaults verbatim, so options
declared as {'parser': 'parse_bool', 'default': 'true'} produced the
string 'true' instead of True. Harmless while defaults never left
database_config.options, but #1062 began forwarding them into DataFrame
adapter constructors: polars-df received rechunk='true' and every table
load failed with "failed to extract field Extract.rechunk" (the three
polars-df e2e failures in the 2026-07-09 release canary). String 'false'
defaults (polars streaming, sqlite check_same_thread) were also truthy,
silently inverting the intended default for truthiness consumers.

PlatformOptionSpec now parses string defaults through the declared
parser at construction, so defaults always carry the same type as
parsed user input. Identity/str-parser specs are unchanged.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/benchbox/core/hooks/platform_hooks.py
+++ b/benchbox/core/hooks/platform_hooks.py
@@ -42,6 +42,16 @@ class PlatformOptionSpec:
     choices: Iterable[Any] | None = None
     aliases: tuple[str, ...] = field(default_factory=tuple)
 
+    def __post_init__(self) -> None:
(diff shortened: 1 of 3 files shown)


joeharris76:
@codex review

---

PR #1043, opened 2026-07-08
Release v0.3.1

joeharris76:
## Release v

This PR was opened by `make release-cut VERSION=X.Y.Z` from `develop`.
It cuts the release branch, bumps the version sources, generates the
CHANGELOG entry from the `origin/main` patch delta, drops maintainer paths
(`_project/`, `_blog/`, agent configs, dev-tooling root files), and queues the
change for a squash-merge into `main`.

### Reviewer checklist

- [ ] CHANGELOG.md has an accurate `## [X.Y.Z] - YYYY-MM-DD` entry
- [ ] `pyproject.toml` and `benchbox/__init__.py` show the new version
- [ ] Documentation landing pages and `landing/` show the new version
- [ ] `_project/`, `_blog/`, agent config dirs, and dev-tooling root files
      were removed from this branch (verify with `git diff main...HEAD --stat`
      — the diff should look like a curated subtree, not arbitrary edits)
- [ ] No surprise file additions (the curation only *removes*)
- [ ] `validate-base` is green for this release branch
- [ ] `release-required-result` is green on this branch
- [ ] Release canary freshness is green (latest `release-canary.yml` run is
      successful, within 48h, and its checked `develop` SHA is an ancestor of
      this release head)
- [ ] Ruleset drift is green in the latest release canary

### Release-required guarantee

`release-required-result` means this branch passed the fast test lane,
credential-free integration-not-slow suite, isolated exact-one-wheel package
install smoke, dependency upper-bound check, and release-branch curation check.
It does not cover live cloud credentials, stress suites, or long-running UAT.
Slow/resource-heavy coverage is enforced by the 48h release canary freshness
gate in `validate-base`, not by rerunning that suite on every release PR.

`release-canary.yml` runs the credential-free non-fast canary
`(slow or resource_heavy) and not (stress or live_integration)` and the
ruleset drift check against `docs/operations/repo-admin-settings.md`.
Scheduled runs execute from the default branch, then check out `develop` and
record the checked SHA in `release-canary-summary.json` for release-readiness.
The first release that introduces the workflow may run the same evidence inline
from `validate-base` until GitHub can run the canary from the default branch.
Emergency override requires repository variables
`RELEASE_READINESS_OVERRIDE_SHA` (exact release head SHA) and
`RELEASE_READINESS_OVERRIDE_REASON` (incident/approval record).

### After release-required checks are green

```bash
make release-finalize VERSION=X.Y.Z
```

`release-finalize` re-checks that the required `validate-base` and
[...]

PR #1089, opened 2026-07-10
fix(release): fold the -s ours history-alignment merge into release-cut

joeharris76:
main and develop have unrelated roots (Phase 4 filter-merge) and diverge
permanently: main carries one squash commit per release that develop never
sees, and the 2026-04-27 amendment removed A4 step 6 (rebase develop onto
main). A vX.Y.Z branch cut cleanly from develop is therefore not mergeable
into main -- GitHub cannot compute a merge ref, so the release PR sits at
CONFLICTING and *no CI runs at all*: neither validate-base nor
release-required-result ever trigger.

Every cut before v0.3.1 (#711, #712, #1029, #1043, #1072) worked around this
by hand. release-cut now performs the merge itself, between the
"Release vX.Y.Z" commit and the push.

-s ours records origin/main as a second parent while preserving the curated
release tree byte-for-byte; release-finalize's squash-merge then collapses
the branch to a single commit, so the merge never reaches main's
release-only ledger. Verified: after the merge, merge-base(origin/main, HEAD)
== origin/main, so GitHub's three-dot PR diff is exactly the release delta.

The recipe guards the merge by asserting the tree hash is unchanged across
it. A -s ours merge cannot change the tree by construction, so the guard is
really an assertion that the strategy is still ours -- any other strategy
drags main-only content back into the curated tree, silently un-curating the
release. The guard is also a no-op-safe check: if origin/main is already an
ancestor, git reports "Already up to date", the tree is unchanged, and the
cut proceeds.

--no-verify: git merge fires pre-merge-commit (not pre-commit). That hook is
not in default_install_hook_types today, so this is defensive -- but curation
has already deleted .pre-commit-config.yaml and the repo:local hook
entrypoints by this point, so installing it later would otherwise break every
cut here.

Ordering is load-bearing: the merge must follow generate_changelog_entry.py
--since-ref origin/main, which needs main to still be a non-ancestor, and
must precede the push, or the PR is born CONFLICTING again. The new
regression test pins all three positions plus the strategy flags.

Coordination with #1027 (rename main -> release): this change uses the literal
origin/main spelling that the rest of the Makefile already uses, so that PR's
textual rename sweep picks these lines up with the existing ones. Expect a
mechanical conflict in the release-cut header comment (step renumbering).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1062,8 +1062,9 @@ RELEASE_REQUIRED_CONTEXTS := validate-base release-required-result
 #   4. $EDITOR opens CHANGELOG.md for hand-curation (skipped if EDITOR unset).
 #   5. Curate: git rm dev-only/deferred paths (per A3 in single-repo-migration.md).
 #   6. Commit "Release v$(VERSION)" (bump + changelog + curation in one squash-friendly commit).
-#   7. Push, open PR vs main.
-#   8. Sweep stale v* branches on origin (option-c lifecycle).
+#   7. Merge origin/main with `-s ours` so the PR is mergeable and CI can run.
+#   8. Push, open PR vs main.
+#   9. Sweep stale v* branches on origin (option-c lifecycle).
 # Pre-conditions: on develop, clean tree.
 # Usage: make release-cut VERSION=X.Y.Z
 release-cut:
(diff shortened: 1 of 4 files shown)


joeharris76:
@codex review

PR #1090, opened 2026-07-10
fix(release): stop nested claude stall, make release-cut resumable, check curation

joeharris76:
`make release-cut` hung twice during the v0.3.1 cut when driven from inside a
Claude Code session: generate_changelog_entry.py shells out to `claude --print`
to summarize commits, and the nested CLI never returned. The 120s timeout could
not save it either — subprocess.run(timeout=...) kills only the direct child and
then re-waits on the captured stdout pipe, which a surviving grandchild still
held open. The cut died mid-flight, leaving the vX.Y.Z branch created, versions
bumped and uv.lock rewritten with no commit.

- Skip summarization inside a Claude Code session (CLAUDECODE /
  CLAUDE_CODE_ENTRYPOINT / CLAUDE_CODE_SESSION_ID), falling straight through to
  _build_raw_changelog. BENCHBOX_CHANGELOG_SUMMARIZE overrides both ways. Run the
  CLI with stdin=DEVNULL in its own process group so the timeout reaps the tree.
- Make release-cut resumable: changelog generation now leaves an existing
  [X.Y.Z] section untouched, so re-running the cut resumes it and preserves a
  section curated in between. Add release-cut-abort to discard a cut instead; it
  refuses once the branch exists on origin.
- Make curation an explicit, checked step. The old gate refused to skip curation
  only when EDITOR was unset AND there was no TTY, which `EDITOR=true` defeated.
  --check-curation now judges the section's own text: verbatim commit subjects
  (trailing PR number), the manual-edit placeholder, or >60 bullets fail the cut.
  The ceiling is calibrated on shipped releases (0.2.1 has 39 curated bullets;
  the raw v0.3.1 delta had 231). RELEASE_ALLOW_RAW_CHANGELOG=1 accepts a raw draft.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1053,40 +1053,80 @@ run-test:
 
 RELEASE_REQUIRED_CONTEXTS := validate-base release-required-result
 
-.PHONY: release-cut release-finalize
+.PHONY: release-cut release-cut-abort release-finalize
 
 # Cut a release branch from develop in one shot:
 #   1. Create v$(VERSION) branch off develop (develop is not modified).
 #   2. On v$(VERSION): bump version sources (scripts/update_version.py).
 #   3. On v$(VERSION): generate CHANGELOG.md entry from the origin/main patch delta.
-#   4. $EDITOR opens CHANGELOG.md for hand-curation (skipped if EDITOR unset).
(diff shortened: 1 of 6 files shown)


joeharris76:
@codex review

PR #1097, opened 2026-07-10
fix(make): stop pr-conflict-scan false-positiving on clean merges

joeharris76:
pr-conflict-scan parsed the deprecated three-arg
`git merge-tree <base> <ours> <theirs>` and grepped for
`changed in both`, `added in both`, `removed in local`, and
`removed in remote`. Those are informational trivial-merge headers
emitted for any path both sides touched -- not conflicts. Because
nearly every PR edits CHANGELOG.md, the probe warned on almost every
pair: 4 of 5 warnings against open PRs were false.

Switch to `git merge-tree --write-tree --name-only` and branch on its
exit status (0 clean, 1 conflict, >1 error). Verify the remote ref
first: a bad ref also exits 1 and would otherwise read as a conflict.
The warning now names the conflicting files.

Anchoring the old grep to `^<<<<<<<` is not a fix -- the legacy format
emits the marker inside a diff body as `+<<<<<<< .our`, so an anchored
grep matches nothing and the scan would go silent on every real
conflict. A test pins that.

Verified against live PRs: silent on the four clean pairs, warns only
on #1090 (fix/release-cut-nested-claude, Makefile), which GitHub
independently reports as CONFLICTING.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1359,8 +1359,16 @@ pr-refresh:
 	$(MAKE) -s pr-open
 
 # Pure-git pairwise textual-conflict probe. Caller passes BRANCH=<current>;
-# we compare HEAD against every other open PR head via `git merge-tree` and
-# print warnings. Warn-only; does not exit non-zero. Used internally by pr-open.
+# we compare HEAD against every other open PR head via `git merge-tree
+# --write-tree` and print warnings. Warn-only; does not exit non-zero. Used
+# internally by pr-open.
+#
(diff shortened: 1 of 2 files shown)


joeharris76:
Superseded by #1098. This PR was opened from the wrong head branch: a concurrent session checked out `feat/explorer-funding-cards` in the shared pool worktree mid-task, so my commit landed on that branch. Same commit, correct branch, in #1098. Leaving the remote branch in place rather than deleting it, since the name belongs to in-flight work.

PR #1106, opened 2026-07-10
fix/throughput timeout leak success gates

joeharris76:
## Summary

Surface leaked timed-out throughput streams, add opt-in cooperative cancellation, and make the TPC-DS success gate configurable with a single source of truth.

Implements TODO `throughput-timeout-leak-and-success-gates`.

## The headline finding: the per-stream timeout was dead code

`StreamRunner.execute()` enforced its per-stream timeout with `future.result(timeout=...)` inside an `as_completed()` loop. But `as_completed()` only yields a future **after it is already `done()`**, and `.result(timeout=...)` on a done future returns or raises immediately without ever consulting the timeout. The `except TimeoutError` branch was therefore **unreachable in real runs**: a normal run silently blocked for as long as the slowest stream took, with **zero timeout enforcement**.

This is a genuine latent bug, deeper than the TODO's framing (which assumed the mechanism worked but couldn't cancel). Fixed by passing the timeout to `as_completed(pending, timeout=...)` itself, plus a fallback loop that classifies any still-pending future as timed-out/leaked.

Leaked streams are now **always** `logger.warning(...)`-ed (previously gated behind `config.verbose`, so they could be entirely silent) and recorded in `result.errors` with an explicit "leaked … cannot forcibly cancel" message.

### Timeout semantics (documented)
`as_completed(pending, timeout=...)` bounds the **whole remaining wait**, not each future. This is equivalent to a per-stream timeout under the invariant that all streams start at ~t0 (`max_workers` defaults to `num_streams`, so nothing queues). The module docstring states this, and a runtime `logger.warning` now fires when `max_workers < num_streams` **and** a timeout is set, because queued streams would share the single t0 deadline and could be flagged timed-out before running a query. Default behavior is unchanged; previously there was no timeout at all, so this is strictly an improvement.

## Opt-in cooperative cancellation
`config.cancel_on_timeout` (default **`False`**) on both TPC-H and TPC-DS configs. When enabled, `execute()` creates one `threading.Event` per stream and sets only the **timed-out stream's own** event; each `_execute_stream` polls its own event once per query and breaks early. No thread is ever hard-killed (per the TODO's anti_pattern). Documented residual: `ThreadPoolExecutor.__exit__` still calls `shutdown(wait=True)`, so with cancellation off (the default) a genuinely hung stream still blocks `execute()` — this matches the TODO's explicit "at minimum this should be surfaced" bar.

## Success gates
- TPC-DS gate is now configurable via `TPCDSThroughputTestConfig.min_success_rate` (default **0.70** preserved), replacing **three** independently hardcoded `0.7` literals — the run-level gate, `_finalize_stream_success()`'s per-stream gate, and a third in `validate_results()` the TODO didn't name. Single source of truth.
- TPC-H default stays **0.99**; its per-stream gate remains the stricter `queries_failed == 0`.
- **Corrected an invented spec citation.** The prior TPC-H comment claimed the spec "allows up to 1% query failures". Both bundled spec PDFs were checked: **neither TPC-H nor TPC-DS defines a partial-success / acceptable-failure-rate allowance** (TPC-H 5.1.1.6 defines a failed run as one that "did not complete successfully due to unforeseen system failures"). Both defaults are now documented as **BenchBox-internal usability tolerances, not TPC compliance gates**, on each config dataclass and in `docs/development/throughput-result-alignment.md`. Sub-100% runs remain ineligible for compliant reporting regardless of the knob.
- Documented that TPC-DS's `min_success_rate` now couples the run-level and per-stream gates (previously two independent literals). Same defaults; a user lowering one now lowers both.

## Review + fix
Opus 4.8 review confirmed the dead-code claim against the original source and stdlib semantics, and confirmed gates are single-source, the spec reframing under-claims safely, and merge-conflict discipline vs #1100 held.

It found one **Required** issue, now fixed: **stale cancel events leaked across runs.** `execute()` set `config._stream_cancel_events` only when cooperative cancel was on and never cleared it, while both specs read the events by *presence* rather than gating on `cancel_on_timeout`. So a config reused across runs (run 1 with `cancel_on_timeout=True` where a stream timed out → its `Event` is `set()`; run 2 with it `False`) would immediately mark that stream cancelled. The TPC-H docstring even asserted an invariant the code didn't guarantee.

Fixed with both layers: `execute()` now resets `_stream_cancel_events = {}` on the non-cooperative path, **and** both specs gate the lookup on `getattr(config, "cancel_on_timeout", False)`, so a stale attribute can never take effect. Docstring corrected.

Three regression tests added (one unit, two integration, incl. one driving the real `TPCHThroughputTest.run()` twice on one config with a genuine timeout). **Proven non-vacuous**: stashing only the source fix while keeping the tests makes all three fail exactly as predicted; restoring it makes them pass.

## Verification
- `tests/unit/core/throughput/` → **18 passed**
- `test_tpch_throughput_test.py` + `test_tpcds_throughput_test.py -m slow` → **31 passed**
- `test_throughput_corrections.py` → **10 passed**
- `test_throughput_real_duckdb.py -m integration` (the oracle) → **5 passed**
- `make lint-imports` → 3 kept, 0 broken · `test_marker_strategy.py` → 8 passed · ruff clean
- Flake risk low: cooperative-cancel behavior is pinned by **deterministic pre-set-event** tests; only the runner-level timeout classification uses real sleeps, with a 10x margin where the sleep is a lower bound.

## Follow-up (non-blocking)
A `shutdown(wait=False)`-style non-blocking return would remove the residual "default-off run still blocks on a hung stream at executor exit". Out of scope here (the TODO explicitly accepts surfacing over killing).

## CI follow-up: robust cancel-event guard
The first push reddened the ubuntu test leg: `tests/unit/tpch/test_tpch_compliance.py::...uses_stream_specific_permutations` failed because it drives `_execute_stream` with a `MagicMock()` config, where `config.cancel_on_timeout` and `config._stream_cancel_events.get(id).is_set()` are all truthy Mocks — so the new cooperative-cancel poll fired spuriously ("stopping after 0/22 queries"). Production configs use real `bool`/`dict`/`threading.Event`, so production was never affected, but the guard must reject malformed state.

Fixed (source only; the out-of-scope compliance test is unmodified): the cancel event is honored only when `config.cancel_on_timeout is True` (a literal bool, not a truthy object), `_stream_cancel_events` is a real `dict`, and the resolved entry is a real `threading.Event` — otherwise it resolves to no-cancel before `is_set()` is consulted. Hardened identically in both TPC-H and TPC-DS paths. Real-Event cooperative-cancel regression tests still pass (default never cancels; a real set Event still cancels).


joeharris76:
@codex review

PR #1290, opened 2026-07-23
test: isolate tuning forwarding from optional drivers

joeharris76:
## Summary

- isolate the registry-wide tuning-forwarding test from optional platform drivers
- keep each adapter's real `from_config()` implementation under test
- bound and verify the constructor-only test patch so it cannot leak across cases

## Root cause

The test called every registered adapter's concrete constructor and converted missing optional-driver errors into skips. That coupled a configuration-forwarding contract to cloud/runtime extras that are intentionally absent from the base development environment.

## Validation

- focused forwarding sweep: 47 passed
- fast lane: 25,528 passed, 15 skipped
- `make pr-preflight`: passed; CI-marker fast lane 25,513 passed, 15 skipped
- Ruff check/format and `ty check`: passed (ty reports existing non-blocking diagnostics)

The remaining skips are unrelated fixture, data, GPU/DataFrame, and isolated-runtime conditions; no optional-driver skips remain in the forwarding sweep.

### DIFF
--- a/tests/unit/platforms/test_tuning_config_forwarding.py
+++ b/tests/unit/platforms/test_tuning_config_forwarding.py
@@ -15,15 +15,9 @@ lands on the instance. It is intentionally parametrized over the live
 platform registry (not a fixed platform list) so a future platform that adds
 ``from_config`` without forwarding fails this test automatically.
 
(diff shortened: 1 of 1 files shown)


joeharris76:
@codex review

PR #1297, opened 2026-07-24
test(dataframe): guard CSV null-semantics skips against regressive un-skip

joeharris76:
The modin/cudf availability skips in the CSV empty-string/null-semantics
test gate a driver-*essential* contract: the adapter performs the CSV load
whose empty-string-vs-NULL result is the assertion. That is the opposite of
PR #1290's from_config() forwarding sweep, which only inspects
constructor-set attributes and so can swap in PlatformAdapter.__init__ to
drop the driver gate.

Add a guard comment at the skip site so a future "un-skip the not-installed
tests" pass does not isolate the constructor here -- which would assert
nothing (or crash on the driver import) rather than test the load. No
behavior change; the skip set is unchanged (16 passed, 4 skipped).


joeharris76:
@codex review

joeharris76:
@codex review

joeharris76:
@codex review

joeharris76:
@codex review

joeharris76:
@codex review

PR #1330, opened 2026-07-29
fix(tests): compare worktree-claim paths on a canonical form

joeharris76:
## What

Three `windows-latest` nightly failures in `tests/integration/worktree/`, two distinct causes.

## 1. Path assertions (2 tests)

The Makefile recipe prints whatever *its own shell* considers the path to be. On Windows that shell is Git Bash/MSYS, which renders `C:\Users\...` as `/c/Users/...`:

```
assert 'WORKTREE_PATH=C:\\Users\\...\\BenchBox.pool-01'
    in 'WORKTREE_PATH=/c/Users/.../BenchBox.pool-01\n'
```

The claim was **correct** — the assertion failed on spelling alone.

Replaced the substring check with an exact comparison on a canonical form. Two deliberate choices:

- **The drive translation is gated on `os.name == "nt"`.** On POSIX a leading single-letter segment (`/a/b`) is an ordinary directory; rewriting it to `A:/b` would be a real bug. Verified `/a/b/c` is untouched on POSIX.
- **`emitted_worktree_path()` asserts there is exactly one `WORKTREE_PATH=` line**, so this is *strictly stronger* than the substring test it replaces, not a loosened one.

## 2. INT trap (1 test)

Platform-gated, not fixed — the behaviour does not exist on Windows. The test:

- shims `git` with an extensionless `#!/bin/sh` script placed first on `PATH`, but Windows resolves `git.exe`/`git.cmd` and never executes an extensionless file, so **the shim is bypassed entirely**;
- drives the recipe's INT trap with `kill -INT $PPID`, which has no Windows equivalent.

The nightly shows exactly that: `returncode 0`, `stderr="Switched to a new branch 'feature/signal-test'"` — real git ran, the claim succeeded cleanly, there was no interrupt to trap, so `assert result.returncode != 0` failed as `assert 0 != 0`. Skipping is the honest outcome; making it "pass" on Windows would mean asserting nothing.

## Verification

The Windows branch cannot run here, so I exercised `normalize_path` directly against the **literal strings from the nightly log**:

```
windows: match = True
  -> c:/users/runneradmin/appdata/local/temp/.../benchbox.pool-01
windows: stdout parse = True
posix  : /a/b/c stays = True
posix  : case preserved = True
```

- `test_worktree_claim_benchbox_scratch.py` + `test_worktree_claim_signal_exit.py` — 4 passed on macOS (the signal test still runs on POSIX)
- `check-scope --strict` — scope OK, 2 changed files
- **Must-preserve holds**: no script under `_project/scripts/` was touched; POSIX behaviour is unchanged. Collected test count is unchanged (the helpers are not test functions), so no fast-lane cap movement.

Windows proof arrives with the next Nightly Validation run.

Closes `nightly-windows-worktree-claim-path-assertions`.

### DIFF
--- a/tests/integration/worktree/test_worktree_claim_benchbox_scratch.py
+++ b/tests/integration/worktree/test_worktree_claim_benchbox_scratch.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import os
+import re
 import subprocess
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

joeharris76:
@codex review

PR #1332, opened 2026-07-29
fix(data-org): detect TBL trailing delimiters under CRLF line endings

joeharris76:
The Windows nightly leg failed 8+ tests in test_sorting.py with
pyarrow.lib.ArrowInvalid "CSV parse error: Expected N columns, got N+1"
on rows like `3|1998-03-01|`.

_source_has_trailing_delimiter() classified framing by seeking to the
last two bytes and testing `tail.endswith((b"|\n", b"|"))`. A CRLF file's
tail is `b"|\r\n"`, which matches neither, so trailing-delimiter framing
was reported as absent. The reader then omitted the synthetic trailing
column and handed PyArrow N column names for N+1 parsed fields.

The test helpers open in text mode, so they emit LF on macOS/Linux and
CRLF on Windows - the same source file, different bytes. That is why this
only ever reproduced on the nightly windows-latest legs.

Replaced the private byte probe with the shared ingest-side helper
benchbox.utils.file_format.has_trailing_delimiter, already used by every
other reader (duckdb, datafusion, polars, and the DataFrame adapters). It
reads in text mode, so universal newlines normalize CRLF before the field
count is taken, and it compares field counts against the schema instead
of matching a byte tail - which also fixes the latent TPC-DS case where a
10-column table whose generator emits 9 values plus a trailing delimiter
was misclassified as needing a dummy column.

The two "regex assertion" failures in the same file were the same root
cause, not message drift: _read_tbl_files() runs before _sort_table(), so
ArrowInvalid was raised before the sort/partition column validation could
raise ValueError. pytest.raises(ValueError, match=...) failed on the
exception type; the messages are identical on every platform. No regex
changes were needed.

Both new tests write explicit CRLF bytes so the framing is exercised on
every platform rather than only on the Windows nightly leg, and both were
confirmed to fail with the exact nightly signature when the production
fix is reverted.

## Review notes — skipped considerations

- **Fast-lane ceiling not bumped.** This PR adds 2 fast tests. `timing_policy_check.py --strict` exits 0 with **0 violations**; it emits only `FAST_LANE_WARNING: headroom 82 below 100`. That warning is **pre-existing** — develop's headroom was 84 before these 2 tests — and `_project/config/**` is outside this item's scope globs (`todo check-scope` passes with exactly 2 changed files). Per the ceiling log's own post-2026-07-24 convention the +500 quantum bump should be a standalone structural entry, not folded into a test-adding PR. Flagged for a separate bump.
- **`_write_tbl`/`_write_tbl_clean` left in text mode.** They emit the platform-native line ending, which is precisely what surfaced this bug on Windows. Converting them to explicit bytes would have removed that coverage, so the new CRLF helper was added alongside rather than replacing them.
- **`_source_has_trailing_delimiter` kept as a wrapper** rather than inlined at its single call site: it still carries the empty-`source_files` guard and the framing rationale.
- **Behavior change beyond the reported failure:** delegating to the shared helper also switches classification from "last line ends with `|`" to "field count exceeds schema width". This is strictly more correct (it fixes the latent TPC-DS 10-column/9-value case the byte probe got wrong) but is a real semantic change worth a reviewer's eye.
- **`_check_first_nonempty_line` swallows read errors** (`except Exception: return False`) where the old probe would propagate. Accepted for consistency with every other reader; `False` matches the bundled binaries' normalized output.
- **No Windows proof yet.** These paths cannot be reproduced on macOS; local rungs are the LF+explicit-CRLF equivalents. Nightly Validation supports `workflow_dispatch` and will be triggered after merge.

### DIFF
--- a/benchbox/core/data_organization/sorting.py
+++ b/benchbox/core/data_organization/sorting.py
@@ -24,6 +24,7 @@ from benchbox.core.data_organization.config import (
     SortColumn,
     SortOrder,
 )
+from benchbox.utils.file_format import has_trailing_delimiter
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


joeharris76 (commented):
The new tests cover CRLF only when a trailing delimiter is present. Please add a clean-CRLF negative case (for example, `1|1992-01-02
` with two declared columns) and assert it is read as exactly two fields. Without that case, a regression where `has_trailing_delimiter(...)` always returns true for CRLF input would still satisfy both added tests, because both fixtures use `_write_tbl_crlf`, which always appends `|` before `
`.

joeharris76:
@codex review

joeharris76:
@codex review

PR #1338, opened 2026-07-29
fix(cli): preserve main submodule identity

joeharris76:
## Summary
- stop `benchbox.cli` from caching the exported `main()` function over the real `benchbox.cli.main` submodule
- make the compatibility submodule callable so `from benchbox.cli import main; main()` remains supported
- preserve generic lazy submodule imports and heavy-dependency isolation
- restore ordinary string patch targets such as `benchbox.cli.main.get_config_manager`

## Root cause
The package-level `__getattr__` special-cased `main`, imported the function from the submodule, and cached that function on the package. Attribute-first access therefore replaced the submodule identity, which breaks Python 3.10-era import and patch resolution. Returning only the submodule fixed resolver behavior but silently removed the historic callable package export, so the compatibility module now uses a small `ModuleType` subclass whose call delegates lazily to its public `main` attribute.

## Verification
- original tracker verification ladder passed, including CLI entrypoint, lazy import, Python 3.10 semantic probe, and the CLI suite
- review-follow-up tracker rungs 1-4 passed: module and callable identity, dotted patching, lazy dependencies, and focused tests
- broader focused regression surface: 23 passed
- Python 3.10.17 compatibility probe passed against the actual CLI package files with a minimal parent-package shim (the standalone interpreter lacks project PyYAML)
- hosted tracker strict scope check passed for all five PR paths
- `make pr-preflight`: all substantive checks passed, including 25,976 fast tests; the final report-only artifact-hygiene guard identified 3.18 MB of pre-existing pool artifacts, which were preserved

## Compatibility
- `benchbox.cli.main` remains a real module for dotted imports, patching, and introspection
- `from benchbox.cli import main` remains callable
- the installed console entrypoint remains `benchbox.cli.app:main`
- importing `benchbox.cli` or `benchbox.cli.main` still does not import Polars or DataFusion

### DIFF
--- a/benchbox/cli/__init__.py
+++ b/benchbox/cli/__init__.py
@@ -19,12 +19,7 @@ __all__ = ["main"]
 
 
 def __getattr__(name: str) -> Any:
-    if name == "main":
-        from benchbox.cli.main import main
-
-        globals()[name] = main
-        return main
-    # Lazy-load any submodule (commands, orchestrator, cloud_storage, etc.)
+    # Lazy-load any submodule (main, commands, orchestrator, cloud_storage, etc.)
     try:
         module = importlib.import_module(f"{__name__}.{name}")
     except ModuleNotFoundError:
--- a/benchbox/cli/main.py
+++ b/benchbox/cli/main.py
@@ -3,6 +3,8 @@
 from __future__ import annotations
 
 import importlib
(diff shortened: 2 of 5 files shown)


joeharris76:
@codex review

PR #1340, opened 2026-07-29
fix(e2e): re-navigate to clear the explorer zero-row snapshot race

joeharris76:
The Chromium blocking gate has been intermittently red for weeks. The two
previous fixes read it as slow DuckDB-WASM cold-start (30s→45s wait, 60s→90s
per test, 1→2 retries). **It is not a latency problem**, so none of that could
have worked — and this PR shows why.

## What the measurements say

Reproduced locally on an idle machine, serial (`--workers=1`), with the current
45s budget in place:

| Observation | Value |
|---|---|
| Wait budget consumed by every flaky spec | the **entire** 45s, element never appeared |
| DuckDB-WASM cold init (suite's own perf spec) | **P50 564ms / P95 978ms** |
| Leaderboard data after init | **P50 6ms** |

A budget that is never nearly met is not the constraint. A healthy load
finishes ~45x inside the old 30s budget.

## Actual root cause

A **cold-snapshot zero-row race**: the first keyed query against a
not-yet-queryable DuckDB-WASM snapshot returns zero rows, the view renders its
empty state, and it never re-queries. No single-shot budget can ever succeed
against that — only a fresh navigation. That is exactly why CI (`retries: 2`)
went green while the same specs kept failing locally (`retries: 0`).

Confirmed three separate ways: in `compare-entrypoints` the route heading
rendered while the result rows stayed empty; same in `responsive` (heatmap) and
`index-sort-headers` (grid rows).

**Second, compounding bug:** `waitForDataLoaded(page, /TPC-H Results/)` waits on
a *shell-rendered route heading*. It is not a data gate at all, so every
row-bound assertion after it raced the snapshot with only the 10s default
`expect` timeout.

## Change

- Data waits run three attempts (10s + 8s + 8s) and **re-navigate to the entry
  URL between them**, each navigation bounded at 10s: ~46s worst case, about
  the 45s it replaces. Attempts are deliberately *shorter* — a racing wait now
  costs ~10s before re-navigating instead of burning 45s, which matters because
  several specs chain data waits inside one 90s per-test cap.
- Re-navigation is **skipped if the page has left the URL the wait began on**,
  so a wait placed after a click cannot rewind the page under a test.
- `waitForResultRows(page, scope, minimum)` gates on real `tbody tr[data-testid]`
  rows **within the table under test** (a page-wide selector would also match
  the excluded-runs disclosure and the heatmap).
- Failure message now reports attempts, budgets, elapsed time and the URL, and
  explicitly says *not* to fix it by widening the budget — the previous message
  reported only the last 12s attempt, which is what invited the original
  misdiagnosis.

Also corrects the prior diagnosis in `docs/operations/browser-ci.md` and the
`playwright.config.ts` comments, including the claim that Chromium runs 2
workers: `test:e2e:chromium` has forced `--workers=1` since #417 (May), so the
"concurrent-worker contention" reading was never right.

## Honest limits

**This is mitigation, not a fix, and it costs one class of coverage.** A
regression that breaks only the *first* load — every user's first visit renders
the empty state, a reload fixes it — now passes on attempt 2. That is called
out in the docs rather than glossed. To keep the coverage somewhere,
`e2e/failures/platform-index-cold-load.spec.ts` is deliberately **left on
single-shot waits** as the dedicated cold-load guard (the pass-1 bug it pins
surfaced as *partial* rows, 1 of 5).

The root cause belongs in `results-explorer/src/**`, which this item lists as
`do_not_modify`. Tracked as `explorer-cold-snapshot-zero-row-race-20260729`
(promoted from a deferral on this item). **Until that lands, residual flake
remains possible at any data-bound assertion still waiting single-shot.**

## Verification

- **3 consecutive clean local runs** (111 passed, `retries: 0` — harsher than
  CI's 2) against a baseline where **2 of 4 runs failed**.
- Suite time **6.3m → 2.2m**, because racing waits no longer burn 45–90s each.
- `npm run typecheck` clean; 820 vitest tests pass; app behavior untouched
  (`results-explorer/src/**` unmodified).
- No assertion was weakened or removed — the two replaced `expect(...)` calls
  run the identical assertion inside the helper.

Note the iteration history here is itself evidence: runs 4 and 9 failed at
*new* sites each time (`responsive`, then `direct-route`), which is what forced
converting the remaining single-shot data waits rather than patching one spec.

Tracker: `explorer-e2e-dataload-flakiness-20260724`

### DIFF
--- a/docs/operations/browser-ci.md
+++ b/docs/operations/browser-ci.md
@@ -11,7 +11,7 @@ and anyone triaging a red or advisory browser-lane check on a PR.
 
 | Browser  | Job name                          | Scope                | Gate       | Status as of 2026-07-23 |
 |----------|------------------------------------|-----------------------|------------|--------------------------|
-| Chromium | `Chromium (full suite, blocking)` | Full `e2e/` suite      | Blocking   | Cold-start flake mitigated 2026-07-25 (see below) |
+| Chromium | `Chromium (full suite, blocking)` | Full `e2e/` suite      | Blocking   | Zero-row snapshot race mitigated 2026-07-29 (see correction below) |
 | Firefox  | `Firefox (@smoke, non-blocking)`  | `@smoke`-tagged subset | Advisory   | Green in recent history |
 | WebKit   | `WebKit (@smoke, non-blocking)`   | `@smoke`-tagged subset | Advisory   | Fixed 2026-07-23 (see below) - was deterministically red on PRs #1264 and #1270 |
 
@@ -41,6 +41,8 @@ below, not a functional break in the PR under test. The required
 `test:e2e:chromium` command runs both Playwright invocations with
 `--workers=1`, so the live cause is serial DuckDB-WASM cold-start exceeding
 the 30s `waitForDataLoaded` budget, not concurrent-worker contention.
+(Superseded - see the 2026-07-29 correction below. It is not a cold-start
+latency problem at all.)
 
 Mitigation (harness-only, no `results-explorer/src/**` change; the blocking
 command remains serial):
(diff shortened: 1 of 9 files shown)


joeharris76:
@codex review

PR #1342, opened 2026-07-29
chore(ci): bump the fast-lane ceiling to 26550

joeharris76:
develop tip collects 26015 fast tests against a 26050 ceiling -- headroom
35, so the next test-adding PR of any size trips FAST_LANE_VIOLATION in
CI's lint job. Bump one +500 quantum per the convention in
_project/config/fast_lane_ceiling_log.md, restoring 535 headroom (above
the 250 floor).

Structural only: no test-count change of its own. Kept standalone rather
than folded into a test-adding branch so the ceiling stays a separate,
reviewable decision -- the 2026-07-29 nightly-Windows batch (#1330,
#1332) shipped without bumping for the same reason.


joeharris76:
@codex review

PR #1343, opened 2026-07-29
feat(explorer): drill down into the per-statement applied-tuning receipt

joeharris76:
#1287 surfaced the tuning verified-**state** in the run receipt, but the
evidence behind that state — the per-statement introspection receipt — rides in
the `.applied.json` companion, which `_project/scripts/explorer_pipeline/` never
read. A reader could see `applied_verified` with no way to see *which*
statements were corroborated against the catalog, or why one was not.

## What this adds

The pipeline ingests `{stem}.applied.json` (mirroring the existing
`.tuning.json` path in `transformer.py`) and carries its `receipt` sub-object
**verbatim** onto the read model as `results.applied_receipt`, canonically
re-serialized so the stored string is deterministic. `RunReceipt.tsx` renders it
in a collapsed `<details>` beneath the existing verification badge: verdict,
kind, table, statement, and expected vs observed columns with the diff or reason
when the platform recorded one.

**Read-only by construction.** The explorer never recomputes a verdict or a
corroboration decision — it displays what the platform recorded at introspection
time.

**Degrades gracefully.** Companion absent, unreadable, malformed JSON, not an
object, no `receipt` key, or no `entries` → no drill-down, no error. Legacy
bundles and runs where introspection never ran render exactly as they do today,
badge included. The parse helper cannot throw.

## Bug fixed in passing

Bundle discovery (`pipeline.py`) excluded `.plans.json` and `.tuning.json` but
**not** `.applied.json` — while `COMPANION_SUFFIXES` has included it since
#1292 and `bundle_publisher` copies it next to every bundle. A published
applied-ledger companion was therefore picked up and transformed as if it were a
result bundle. Discovery now gates on the canonical `COMPANION_SUFFIXES` tuple,
so a future companion kind cannot reintroduce the same defect. Covered by a
regression test.

## Contract version is pinned in FOUR places, not three

The read model gains a column, so the contract goes **5 → 6**. Worth recording
for the next person, because the house note says three files:

| File | Caught by |
|---|---|
| `_project/scripts/explorer_pipeline/contract.py` | unit tests |
| `results-explorer/src/db.ts` | vitest |
| `results-explorer/scripts/explorer-build-contract.mjs` | **Chromium e2e only** |
| `tests/unit/scripts/test_explorer_build_contract.py` | **full fast lane only** |

The fourth was missed by a scoped `pytest tests/unit/scripts/explorer_pipeline`
run and only surfaced under the full `-m fast` lane.

## Verification

- `pytest tests/unit/scripts/explorer_pipeline -q` → **304 passed**
- `npx vitest run src/components/__tests__/RunReceipt.test.tsx` → **29 passed**
- `npm test` → **832 passed**; `npm run typecheck` clean
- Full fast lane → **26009 passed**
- Chromium e2e run to exercise the `.mjs` mirror: `validated explorer build
  contract 6`, 109 passed. One unrelated failure —
  `responsive.spec.ts` `waitForDataLoaded` at a 45s budget, i.e. the
  pre-existing cold-snapshot zero-row flake this branch predates the fix for
  (#1340).

## Notes for the reviewer

- **Scope deviation, flagged.** The item scopes edits to
  `_project/scripts/explorer_pipeline/**`, `results-explorer/src/**`,
  `tests/unit/scripts/explorer_pipeline/**`. Four files fall outside and are
  unavoidable: `explorer-build-contract.mjs` and
  `tests/unit/scripts/test_explorer_build_contract.py` are contract-version
  pins that must move in lockstep, and `_project/config/fast_test_lane_policy.json`
  + `fast_lane_ceiling_log.md` are the CI fast-lane ceiling (26050 → 26550 under
  the documented +500 quantum convention; the branch collected 26025 against the
  old cap, headroom 25).
- `benchbox/core/tuning/**` and `results-explorer/src/lib/facetMatching.ts`
  untouched; `matchesFacetKey` / `FACET_KEYS` / `physicalRenderingIdsMatch`
  not touched.
- `applied_receipt` is deliberately **not** added to the list projection, only
  the detail one — it is a potentially large blob and no list surface needs it.

Tracker: `explorer-applied-receipt-drilldown-20260724`


joeharris76:
@codex review

PR #1344, opened 2026-07-29
docs(explorer): sync browser schema doc with the built DDL, and pin it there

joeharris76:
`docs/development/browser-duckdb-schema.sql` documents the browser DuckDB
store, but nothing executed or checked it — so it quietly fell behind
`duckdb_builder.py`.

**The gap was wider than the TODO reported.** Comparing the doc against the
schema `build_full()` actually produces, the `results` table and the
`result_detail_metrics` view were each missing **six** columns, not the two
ADR-2 ones:

| Missing column | Landed in |
|---|---|
| `requested_config_hash` | ADR-1 tuning identities |
| `applied_ledger_hash` | ADR-1 tuning identities |
| `tuning_validation_status` | ADR-1 verified-state (read model v5) |
| `tuning_policy_generation` | ADR-3 policy-generation seam |
| `physical_mechanisms` | ADR-2 §3 |
| `physical_rendering_id` | ADR-2 §3 (read model v4) |

It had been drifting across four read-model versions.

## Change

All six added to both the table and the view, in the builder's declaration
order and carrying the builder's own comments. `duckdb_builder.py` is untouched
— it is the source of truth, and the doc is aligned to it, not the reverse.

## Drift guard (the stretch goal)

`TestSchemaDocMatchesBuiltSchema` compares the doc against the schema the
pipeline actually builds, reusing the existing module-scoped `db_path` fixture.
Order-sensitive equality in **both** directions: a new column that lands
undocumented fails, and so does a column the doc invents.

The pre-existing contract test could not have caught this — it asserts a
hardcoded column set is a `issubset` of the real schema, and never reads the
doc at all.

**Verified the guard actually fires**: deleting `physical_rendering_id` from
the doc fails with `undocumented columns: ['physical_rendering_id']` (and the
matching view assertion), and both pass again once restored.

## Verification

- Item ladder: `grep -c 'physical_rendering_id\|physical_mechanisms'` → **4**
  (expected ≥ 4)
- `pytest tests/unit/scripts/explorer_pipeline -q` → **295 passed**
- Full fast lane → **26000 passed**
- `todo check-scope` → **scope OK** (2 changed files, both in the allowlist)

## Sequencing note

PR #1343 (applied-receipt drill-down) adds a further `results.applied_receipt`
column. Whichever of the two merges second must add that column to this doc, or
the new guard will correctly fail. I'll keep #1343 consistent with whatever
lands first.

Tracker: `docs-browser-duckdb-schema-column-drift-20260724`

### DIFF
--- a/docs/development/browser-duckdb-schema.sql
+++ b/docs/development/browser-duckdb-schema.sql
@@ -74,6 +74,19 @@ CREATE TABLE IF NOT EXISTS results (
     execution_mode       VARCHAR,
     tuning_mode          VARCHAR,
     tuning_hash          VARCHAR,
+    -- ADR-1 bundle-emitted tuning identities (display-only, never a join/dedup
+    -- key): canonical requested-config hash and the physical applied-ledger
+    -- hash. NULL for legacy bundles.
+    requested_config_hash VARCHAR,
+    applied_ledger_hash   VARCHAR,
+    -- ADR-1 tuning verified-state (not_applicable / noop / applied_unverified
+    -- / applied_verified / failed). NULL for legacy bundles; distinct from the
+    -- run/query validation_status column below.
+    tuning_validation_status VARCHAR,
+    -- ADR-3 seam: explicit tuning-policy generation marker (display-only,
+    -- never a join/dedup key). NULL for legacy bundles, treated downstream as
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1347, opened 2026-07-29
feat/findings schema v4

joeharris76:
- **feat(tracker): schema v4 — lossless legacy homes and a delete-safe link FK**
- **feat(tracker): land every legacy capture field losslessly at v4**
- **docs(tracker): v3 -> v4 hosted migration runbook and landed mapping status**


joeharris76:
@codex review

PR #1351, opened 2026-07-30
feat(tracker): phase-5 corpus importer with a falsifiable parity report

joeharris76:
Adds `todo finding import [--corpus DIR] [--dry-run] [--report]` plus the parity
report the phase-5 gate is defined against.

Deliberately NOT sync_drafts: that path renames each landed file to `*.synced`,
and phase 5 must leave the tracked records untouched — the freeze window is both
the requirement-2 proof and the rollback path, and phase 6 deletes only after its
own gates. The importer reads the corpus and nothing else; a regression test
asserts no file is renamed, deleted, or added.

Triage-log prose imports as one `imported_triage_log` event per non-empty line,
verbatim. Parsing it into structured disposition events stays rejected: that
parity claim is either trivial (a blob) or unachievable (prose parsing), and
structured guesses corrupt an audit trail.

The report is falsifiable rather than reassuring — body columns are hash-compared,
every frontmatter-derived column is diffed, extra sections are compared by
position/heading/text, and triage-log line counts must equal stored event counts.
`ok` is true only when every diff list is empty, and the CLI exits 1 otherwise so
the rung cannot pass on a failing report. Tests prove it fails closed on a lost
section and on a lost triage line, not merely that it passes today.

Import is idempotent by record id and records `imported_from` provenance, so a
re-run after a partial import neither duplicates triage events nor re-inserts.

### DIFF
--- a/_project/scripts/todo_findings.py
+++ b/_project/scripts/todo_findings.py
@@ -25,6 +25,7 @@ credentialed landing step (review-protocol §4).
 from __future__ import annotations
 
 import argparse
+import hashlib
 import json
 import os
 import re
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1352, opened 2026-07-30
fix(windows): resolve the largest Nightly Validation failure families

joeharris76:
# Pull Request

## Description

Attacks the `Nightly Validation` Windows failures (56 failed in each of
`test (windows-latest, 3.10)` and `test (windows-latest, 3.12)`; baseline run
`30473819253`).

**Method.** Windows cannot be run in this environment, so rather than fix blind I
reproduced each Windows-only condition *on Linux* first, confirmed the failure count
matched CI, then fixed and re-ran. CI then confirmed the result on real Windows.

### Confirmed by CI

Run [30542333194](https://github.com/joeharris76/BenchBox/actions/runs/30542333194)
on the first commit alone:

| Job | Baseline | After commit 1 |
|---|---|---|
| `test (windows-latest, 3.12)` | 56 failed | **29 failed** |
| `test (windows-latest, 3.10)` | 56 failed | **28 failed** |

That is the 27 predicted from the two families below, landing exactly. Runs for the
second and third commits are queued.

### Family 1 — chDB probe (17 failures)

`ClickHouseAdapter.__init__` probes for chDB eagerly and raises `ImportError` when it
is absent, even though nothing needs the package until a connection opens. chDB ships
no Windows wheels — `pyproject.toml` constrains it to `sys_platform != 'win32'` — so
every test in `test_clickhouse_introspection.py` and `test_clickhouse_plan_capture.py`
died at construction. None of them touches chDB: they drive DDL rendering and
plan/introspection logic against fake connections.

Fixed by isolating the probe, **not** by skipping: a `chdb_probe_satisfied` fixture
satisfies only that lookup, opted into per module via `usefixtures`. Keyed on probe
availability rather than `sys.platform`, so it stays honest if chDB ever ships Windows
wheels and provably does not over-skip.

> Simulated chDB absence on Linux: **17 failed**, matching CI exactly. After:
> **29 passed with chDB present and 29 with it absent, 0 skipped either way.**

The tracker item `windows-clickhouse-local-tests-without-chdb` says "9 failures"; the
observed count is 17.

### Family 2 — `bash` resolves to the WSL stub (11 failures)

`subprocess.run(["bash", "-c", ...])` resolved to `C:\Windows\System32\bash.exe` — the
**WSL launcher stub**, which precedes Git Bash on the runner's PATH. With no distro
installed it prints its usage text in **UTF-16LE** and exits 1, so assertions compared
expected substrings against mojibake:

```
assert 'content-guard=failure' in "W\x00i\x00n\x00d\x00o\x00w\x00s\x00 \x00S\x00u\x00b\x00s\x00y\x00s\x00t\x00e\x00m\x00 ..."
```

which decodes to *"Windows Subsystem for Linux … '&lt;Distro&gt;' to install."* This reads
like a missing-`encoding` bug and was previously triaged as one — but the interpreter is
wrong, so no decoding change can fix it.

`tests/utilities/posix_shell.py` resolves a shell by **probing** candidates with the
bash features these fragments actually use (`set -o pipefail`, a herestring), so it
rejects the stub wherever it lives — and correctly rejects a POSIX-only `/bin/sh`
(dash), which cannot run these `shell: bash` fragments either. It scans *every* PATH
entry rather than taking `shutil.which`'s first hit, because a rejected stub earlier on
PATH would otherwise hide the real bash later on it — the runner's exact layout.

Guards are **per-test** via `skip_without_posix_shell()`, not module-level: 6 of the 16
tests in the classifier module assert on YAML and regexes, need no shell, and must keep
running on Windows.

Also fixes a **scope gap**: `test_validate_submission_changed_bundles.py` hits the same
stub but was absent from that tracker item's `only_modify` list.

### Family 3 — locale-dependent reads (2 product defects)

`Path.read_text()` with no encoding uses `locale.getpreferredencoding()` — cp1252 on
Windows — raising `UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f`. Fixed
in the affected test readers and in the **two product call sites** that had the same
latent bug with no test covering them: `benchbox/utils/dependencies.py` (pyproject.toml,
UTF-8 by PEP 518) and `benchbox/core/tpcds_obt/query_conversion.py`.

### Family 4 — `is_snowflake_stage_path` (product defect)

The function accepts `Union[str, Path]`, but the stage grammar requires `/` after the
stage token and `str(WindowsPath("@~/data"))` is `"@~\data"`. A `Path`-typed Snowflake
stage reference therefore classified as **not a stage** on Windows — and for this
predicate that means a stage path can be treated as local.

While there I widened `isinstance(path, Path)` to `PurePath`: the pure flavours are not
`Path` subclasses, so `PureWindowsPath` input previously fell through to the non-str
rejection. Found because my first regression test used `PureWindowsPath` and still
failed — it was exercising a different branch than production.

### Family 5 — TPC-H CRLF normalization (5 failures, product defect)

`_has_trailing_delimiter` probed the last **two** bytes for `b"|\n"`, but under CRLF the
sequence is `b"|\r\n"` — so the probe only ever saw `b"\r\n"`, missed the delimiter and
reported the file already clean. The streaming rewrite had the matching gap: it replaced
only `b"|\n"`, and its carry held back a chunk-final `b"|"` but not `b"|\r"`, corrupting
rows when a terminator split across a chunk boundary.

This is precisely the case the normalizer exists for — a dbgen built on a platform
outside the precompiled matrix, which on Windows emits CRLF. #1332 intended this support;
the probe width kept it from working.

### Family 6 — bundle export newline determinism (product defect)

`ResultExporter._write_file` opened a text handle without `newline=""`, so
universal-newline translation rewrote `\n` to `\r\n` on Windows. Bundles are
**byte**-defined (`canonical_json_bytes`) and their bytes feed the `companion_hashes`
submission contract, so the same logical bundle exported on Windows would not byte-match
one exported on Linux or macOS.

### Family 7 — POSIX-only APIs

- `test_kill_process_group_uses_killpg_when_available` used `monkeypatch.setattr` without
  `raising=False`, so it errored on win32 where `os.killpg` does not exist. The product
  *already* has the `hasattr` guard. `raising=False` injects the attribute so the POSIX
  ladder is still exercised on Windows; a `skipif` would have dropped real coverage.
- `test_shared_predicate_script_is_executable_for_workflow` asserted a POSIX exec bit via
  `stat().st_mode`, which Windows lacks. It now asserts the mode **git records**
  (`100755`) via `git ls-files --stage` — what actually governs the Linux runner, and
  identical on every platform. Stronger, not platform-weakened.

### Family 8 — path-separator assertions

`tests/uat/test_preflight_budget.py` and `test_runner.py` hardcoded `/tmp` and
`benchmark_runs/results/` while the code under test renders `str(Path)`. Both now derive
the expectation from the same `Path` the code receives — asserting at least as much, and
matching the idiom already in `test_preflight.py`. **No assertion was loosened.**

## Type of Change

- [x] Bug fix

## Testing

Windows verification is by CI. Run `30542333194` confirmed commit 1 (**56 → 29/28**);
runs for commits 2 and 3 are queued. **This PR should not merge until those report.**

Linux verification, simulating each Windows condition:

| Condition simulated | Before | After |
|---|---|---|
| chDB absent (`find_spec` patched) | 17 failed | 29 passed, 0 skipped |
| WSL stub first on PATH | 10 failed | 89 passed |
| `PureWindowsPath` stage reference | failed | 67 passed |
| CRLF `.tbl` input (bytes written directly) | 4 failed | 11 passed |

Every new test was checked against the pre-fix code so none is vacuous.

```
tests/unit/workflows/ tests/unit/test_release_infrastructure.py
  tests/unit/platforms/test_clickhouse_*.py tests/unit/test_posix_shell_resolver.py   207 passed
tests/uat/... tests/unit/utils/... tests/unit/cli/... tests/unit/query_plans/...       247 passed
tests/unit/test_results_exporter.py tests/unit/core/tpch/... tests/unit/test_auto_merge... 56 passed
uv run -- ruff check / ruff format --check (all changed files)                         clean
```

One failure in the broader `-k "exporter or canonical or bundle"` selection,
`test_export_with_invalid_output_dir`, is **pre-existing and unrelated** — verified by
stashing the change: it fails identically without it, because this container runs as
root and can create the supposedly-uncreatable path.

## Public Contract Check

- [x] I updated `docs/reference/public-contracts.md`, or this PR does not change
      public, beta-public, deprecated, generated, experimental, or repo-only
      contract surfaces.
- [x] I checked result schema fields, MCP tool parameters, platform support
      status, wrapper facades, adapter subclassing hooks, and generated docs for
      contract-map impact.
- [x] Any platform/benchmark count claim in docs is generated/checked from
      registry metadata, or explicitly marked editorial/non-authoritative.

Two surfaces worth a reviewer's eye, neither a signature change:
`is_snowflake_stage_path` now honours the `Path` half of its own annotation on Windows
and accepts `PurePath`; `ResultExporter` output is now byte-identical across platforms
(previously Windows-only CRLF). Both make behaviour *more* conformant, and the bundle
change only affects Windows-produced bundles, which could not have matched their hashes
before.

## Documentation

- [x] I updated the relevant user-facing docs.
- [x] I added regression notes for behavior changes.
- [x] I confirmed API contract wording remains accurate.

Rationale is recorded where maintainers will meet it: the `posix_shell` module
docstring, the `chdb_probe_satisfied` fixture docstring, and the new tests' docstrings.

## Code Quality

- [x] I ran formatting and lint/type checks.
- [x] I reviewed impacted modules for backwards compatibility and migration impact.
- [x] I added/updated tests for behavior changes and error paths.
- [x] I confirmed public contracts and release checklists are still valid.

## Artifact Hygiene

- [x] I did not commit raw screenshots, browser reports, generated logs,
      benchmark outputs, or temporary binary artifacts.
- [x] Any committed binary/raw evidence file has durable repo value and is
      justified here with its consumer and size:

None. The reproduction harnesses (a `find_spec` plugin, a WSL-stub script) were
deliberately not committed; the recipes are documented in docstrings instead.

## Notes

**Still red — not claimed fixed.** Roughly 11 of the original 56 remain, left with their
evidence rather than guessed at:

- `tests/uat/test_phases.py` (4) + `test_orchestrator.py` (2) + `test_docker_assets.py`
  (1) — docker compose command/path assertions. **`test_disk_floor_abort_carries_startup_failed_count`
  is a genuine logic bug** (`startup_failed_count` is 0, expected 1), not a path artifact,
  and deserves its own investigation.
- `tests/unit/tpcds/test_dsqgen_streams.py` (3) — the Windows dsqgen binary fails with
  `Runtime ERROR: Distribution over-run/under-run … cities`, which looks like a real
  binary/data problem rather than a test issue.
- `tests/unit/scripts/test_pr_review_followups.py` (1) — the test writes a
  bash-shebang `codex` shim with no Windows-resolvable extension, so `shutil.which`
  never finds it. Fixable with a `.cmd` wrapper over the existing `posix_shell()`
  helper, but fiddly and unverifiable from here.

**Also seen, not caused by this PR:** `test (macos-latest, 3.12)` failed in run
`30542333194` with 22 `InsufficientMemoryError` failures ("~2.0 GB required, but only
2.4 GB available") in `tests/unit/platforms/dataframe/`. macOS was green in the baseline
and nothing here touches memory estimation, so this looks like runner resource pressure —
but it is worth watching rather than assuming.

**Reviewer focus:** the `posix_shell` probe rejecting `/bin/sh` is deliberate — these
fragments are `shell: bash` in the workflows and would fail confusingly under dash. If
you would rather they degrade to `sh`, that is a one-line change to `_PROBE_SCRIPT`.

### DIFF
--- a/benchbox/core/tpcds_obt/query_conversion.py
+++ b/benchbox/core/tpcds_obt/query_conversion.py
@@ -158,7 +158,9 @@ class TemplateLoader:
         if not self.path.exists():
             raise FileNotFoundError(f"Template for query {query_id} not found at {self.path}")
 
-        self.raw_text = self.path.read_text()
+        # Explicit encoding: the read_text default is locale-dependent (cp1252 on
+        # Windows), which mis-decodes any non-ASCII byte in a query file.
+        self.raw_text = self.path.read_text(encoding="utf-8")
         self.definitions: dict[str, str] = {}
         self.body_sql: str = ""
         self.parameters: dict[str, TemplateParameter] = {}
--- a/benchbox/utils/cloud_storage.py
+++ b/benchbox/utils/cloud_storage.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 import logging
 import os
 import re
-from pathlib import Path
+from pathlib import Path, PurePath
 from typing import Any, Iterator, List, NamedTuple, Protocol, Union, runtime_checkable
 from urllib.parse import urlparse
 
@@ -413,8 +413,17 @@ def _match_snowflake_stage(path: Union[str, Path]) -> Union[re.Match, None]:
     Shared by the predicate and by :func:`get_cloud_path_info` so the grammar
     and the stage/sub-path split can never drift apart.
     """
-    if isinstance(path, Path):
-        path = str(path)
+    # PurePath, not Path: this must also cover the pure flavours (PureWindowsPath),
+    # which are not Path subclasses and previously fell through to the non-str
+    # rejection below.
+    if isinstance(path, PurePath):
+        # A stage reference is Snowflake URI grammar, not a local filesystem path, so
+        # it is always "/"-separated. str(WindowsPath("@~/data")) is "@~\\data", which
+        # the grammar rejects because it requires "/" after the stage token -- so a
+        # Path-typed stage silently failed to classify on Windows. Normalize it to
+        # behave like its string form everywhere. Non-stage inputs are unaffected:
+        # they do not start with "@" and never matched either way.
+        path = str(path).replace("\\", "/")
 
     if not isinstance(path, str):
         return None
--- a/benchbox/utils/dependencies.py
+++ b/benchbox/utils/dependencies.py
@@ -34,7 +34,9 @@ def is_development_install() -> bool:
     if (project_root / "pyproject.toml").exists():
         # Verify it's actually the benchbox project
         try:
-            content = (project_root / "pyproject.toml").read_text()
+            # pyproject.toml is UTF-8 by specification (PEP 518); the read_text
+            # default is locale-dependent and cp1252 on Windows.
+            content = (project_root / "pyproject.toml").read_text(encoding="utf-8")
             if 'name = "benchbox"' in content:
                 return True
         except Exception:
--- a/tests/uat/test_no_cli_surface_drift.py
+++ b/tests/uat/test_no_cli_surface_drift.py
@@ -248,7 +248,8 @@ def _source_at_ref(ref: str, path: str) -> str:
 
 def _source_at_worktree(path: str) -> str:
     try:
-        return (REPO_ROOT / path).read_text()
+        # Explicit encoding: the default is locale-dependent (cp1252 on Windows).
+        return (REPO_ROOT / path).read_text(encoding="utf-8")
     except FileNotFoundError:
         return ""
 
--- a/tests/uat/test_preflight_budget.py
+++ b/tests/uat/test_preflight_budget.py
@@ -97,8 +97,12 @@ def test_estimate_largest_scale_peak_disk_uses_largest_configured_scale(tmp_path
 
 def test_disk_headroom_gate_reports_short_root(tmp_path: Path):
     budget = DiskBudget(cells=1, est_peak_gib=10.0, est_steady_gib=8.0, unknown_cells=())
+    # Render the expectation from the same Path the formatter receives: str(Path) uses
+    # the platform separator, so a hardcoded "/tmp" fails on Windows ("\tmp") while
+    # asserting nothing extra. Mirrors the idiom already used in test_preflight.py.
+    tmp_root = Path("/tmp")
     roots = (
-        DiskRootFreeSpace("tmp", Path("/tmp"), 4.0),
+        DiskRootFreeSpace("tmp", tmp_root, 4.0),
         DiskRootFreeSpace("output", tmp_path, 20.0),
     )
 
@@ -107,4 +111,4 @@ def test_disk_headroom_gate_reports_short_root(tmp_path: Path):
     assert check.required_gib == pytest.approx(10.0)
     assert len(check.shortfalls) == 1
     assert check.shortfalls[0].label == "tmp"
-    assert "tmp /tmp: 4.0 GiB free < 10.0 GiB required" in format_disk_headroom_failure(check)
+    assert f"tmp {tmp_root}: 4.0 GiB free < 10.0 GiB required" in format_disk_headroom_failure(check)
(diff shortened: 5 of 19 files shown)


joeharris76:
@codex review

PR #1366, opened 2026-07-30
fix(tpch): make trailing-delimiter normalization CRLF-safe

joeharris76:
## Why

Five Windows nightly failures in `test_tpch_trailing_delimiter_normalization.py` — **two independent bugs**, and the second would have kept the file broken even after fixing the first.

**1. Detection.** `_has_trailing_delimiter` read the last two bytes and matched `b"|\n"`. A CRLF file never matches — its tail is `b"|\r\n"` — so a file produced on Windows was classified *already-clean* and never rewritten. Now reads three bytes and strips the terminator before looking for the delimiter, the same shape as the `sorting.py` fix in #1332.

**2. Rewrite.** The streaming rewrite replaced only `b"|\n"`. Even a correctly detected CRLF file would have kept every delimiter. It now replaces `b"|\r\n"` and `b"|\n"`, preserving whichever terminator it found.

**3. Chunk carry.** A chunk boundary can fall *inside* `|\r\n`. The one-byte carry splits the `|` from its terminator, so the delimiter survives into the output. Carry is now two bytes.

## Why the tests passed locally

The existing fixtures use `write_text`, which emits the platform's native terminator — LF on macOS/Linux, CRLF on Windows. So all five regressions were green locally and red only on the `windows-latest` leg.

The six new fixtures are written as **explicit bytes**, so both framings are exercised everywhere:

- detect and strip under CRLF
- interior empty fields preserved
- **already-clean CRLF is byte-stable** — the must-preserve, including that terminators are not silently rewritten to LF
- idempotence
- an unterminated `|\r` final row
- a chunk-boundary test sweeping **every** chunk size from 1 to the payload length

## Proven to fail, not just pass

| Reverted | Result |
|---|---|
| both fixes | 5 of 6 CRLF tests fail |
| only the two-byte carry | exactly the chunk-boundary test fails |

Each fix is independently guarded.

## Scope note

`benchbox/utils/file_format.py` is in this item's `only_modify` list but needed **no change** — both bugs were in the generator. `check-scope --strict` — scope OK, 2 files.

`tests/unit/core/tpch/` — 295 passed. Fast lane 26085 / 26550.

Closes `windows-tpch-trailing-delimiter-crlf-normalization`.

### DIFF
--- a/benchbox/core/tpch/generator.py
+++ b/benchbox/core/tpch/generator.py
@@ -55,14 +55,21 @@ def _has_trailing_delimiter(path: Path) -> bool:
     dbgen row framing is decided at compile time (EOL_HANDLING), so every row
     in a file shares the same framing; inspecting the file tail is sufficient
     to classify the whole file.
+
+    The line terminator is an independent variable: a file produced on Windows
+    ends ``|\r\n``, not ``|\n``. Strip the terminator before looking for the
+    delimiter rather than matching one spelling of it, or a CRLF file reads as
+    already-clean and is never normalized.
     """
     size = path.stat().st_size
     if size == 0:
         return False
+    # 3 bytes covers the longest terminator ("|\r\n"); the clamped seek handles
+    # files shorter than that.
     with path.open("rb") as handle:
-        handle.seek(max(0, size - 2))
+        handle.seek(max(0, size - 3))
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1367, opened 2026-07-30
fix(ducklake): a stale local catalog file must not skip the postgres data load

joeharris76:
Closes `ducklake-postgres-reuse-keys-off-local-metadata-path`. **Blocks any real
postgres-catalog benchmark run** — found by the first TPC-H SF=1 attempt.

## The failure

```
✅ Database being reused - skipping schema creation and data loading
❌ Data validation failed - benchmark execution halted
⚠️  Empty tables detected: region, nation, supplier, part, partsupp, customer, ...
```

Against a **genuinely fresh** PostgreSQL catalog (`ducklake_table` empty).

## Cause

`handle_existing_database()` decides reuse from `self.metadata_path.exists()`.
For `catalog=postgres` that path describes no part of the catalog — but it is
still always *computed* to a default. So a leftover `.ducklake` file from an
earlier **duckdb-catalog** run at the same benchmark/scale read as "database
exists, reuse it", the runner skipped schema creation and data loading, and
validation then correctly failed against an empty PG catalog.

Reproduced directly:

```
metadata_path exists: True
database_was_reused : True     # on a fresh PG catalog
```

**This is a gap in my own #1358.** That PR added post-ATTACH reuse resolution
for postgres (`_resolve_postgres_catalog_reuse`) but did not stop the
pre-connection check from firing — so the local file still won, and won in the
wrong direction. Its docstring also asserted that a postgres catalog "never has"
a local metadata file, which is simply false.

## Fix

`handle_existing_database()` now declines to decide for the postgres backend and
leaves it to `_resolve_postgres_catalog_reuse`, which inspects the attached
catalog. `duckdb`/`sqlite` are untouched, pinned by a must-preserve test. The
stale file is deliberately not deleted — it is not this backend's to remove.

## Verification

- **The same SF=1 run now passes**: data validation passed, Power@Size
  72118.63 / 76298.44 / 85338.69 across three streams.
- `tests/unit/platforms/test_ducklake_adapter.py` — 107 passed (2 new).
- Mutation: removing the early-return fails the new regression test.
- `check-scope` OK (2 files); ruff/format clean.

## Note on how this was found

Only a real scale run surfaced it. The unit tests for #1358 passed, the live
smoke test passed, and the defect still sat there — it needs a *pre-existing
local catalog file at the default path* plus a postgres backend, which no
hermetic test constructed. That combination is exactly what an ordinary second
run on a developer machine produces.

### DIFF
--- a/benchbox/platforms/ducklake.py
+++ b/benchbox/platforms/ducklake.py
@@ -563,15 +563,30 @@ class DuckLakeAdapter(DuckDBAdapter):
           data loading are skipped and queries run against the existing catalog.
           Pass ``--force`` to wipe the catalog + data dir for a clean rebuild.
 
-        Postgres catalog: this hook runs BEFORE any connection exists and keys
-        off the local ``metadata_path`` file, which a server-side catalog never
-        has - so it can only ever take the "does not exist yet" branch here.
-        Reuse/force for ``catalog == "postgres"`` is therefore resolved after
-        the ATTACH instead, in ``_resolve_postgres_catalog_reuse()``, which
-        inspects the attached catalog itself; the framework reads
-        ``database_was_reused`` after ``create_connection()`` returns, so
-        deciding it there is equivalent for the runner.
+        Postgres catalog: this hook runs BEFORE any connection exists, and the
+        only thing it can inspect is the local ``metadata_path`` - which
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1410, opened 2026-08-01
test(windows): inject unreadable file capability

joeharris76:


joeharris76:
Closing as superseded by #1393, which already replaces this chmod-based unreadable-file test with an injected PermissionError and adds adjacent compressed-read coverage. Keeping one authoritative implementation avoids a same-hunk conflict.

PR #1512, opened 2026-08-04
fix(results): make public anonymization reach a fixed point

joeharris76:
## Problem

`AnonymizationManager.anonymize_result_payload` was not idempotent.
`_hash_public_identifier` hashed whatever it was handed, with no notion that the
value might already be a pseudonym it produced. Curated bundles are stored
already-anonymized, so the Explorer publication boundary hashed them a second
time.

Measured on `develop` across the whole corpus, before the fix:

```
source machine_id    : machine_3d46427037d2
published machine_id : machine_f8dd3066dcd1
IDEMPOTENT? False   (a publish pass changed bytes for 203/203 bundles)
```

Nothing failed — the grouping was just wrong. A curated bundle reached the
Explorer as a two-pass pseudonym, while a fresh community submission from the
**same physical machine** arrived as a one-pass pseudonym. One machine, two
public `machine_id` values. Same for `host_`, `path_`, and every other
identifier family. Pseudonym stability is the whole basis for correlating
results from one machine, so this silently broke it.

## Fix

A value already carrying its own prefix's pseudonym shape now passes through, so
anonymization reaches a fixed point.

The check is scoped to **the prefix being applied** and to **the exact emitted
digest width**, rather than to pseudonym shape in general:

- a capture-side `machine_<16 hex>` is still hashed, so the internal machine id
  is never published verbatim — internal and public identity stay decoupled;
- a `host_` token cannot survive verbatim inside a `path_` field.

This is deliberately tighter than the `^(?:machine|host|path|...)_[0-9a-f]{8,64}$`
pattern that existed in the migration tool, which would have done both of those
things. It also needs no prefix list: it derives from the prefix in hand, and
there are 27 prefixes in use versus the 10 that pattern enumerated.

**Security note.** A submitter can still hand-craft a value of this shape to pick
their own pseudonym. Pseudonym identity is therefore a grouping key only — never
a provenance or trust signal. Trust labels carry provenance. This is documented
on the method.

## Verification

- Publication is idempotent for **203/203** corpus bundles (was 0/203).
- For all **206** pre-#1467 originals, a fresh submission and the curated bundle
  now publish the **same** `machine_id` — the cross-corpus grouping bug is gone.
- `find_public_path_leaks` reports **zero** leaks over both corpora at one and
  two passes.
- Full fast lane: 26621 passed, 0 failures.
- Negative control: reverting the pass-through fails 5 of the new gates,
  including the corpus-wide one.

## On re-deriving the corpus

Not needed, and not done. The premise that the stored corpus is two passes from
raw does not hold: for **206/206** pre-#1467 originals,
`H(raw₁₆) == the current stored 12-hex value`. The corpus was always exactly one
pass from raw — only *publication* was doubling it. Fixing the boundary is
sufficient, and the corpus is untouched.

Published bytes do change, so every `result_id` changes. That is free right now:
`origin/release` carries zero bundles (verified).

## Also

Drops `_preserve_public_hashes` from `results_explorer_corpus_migrate.py`. It was
a looser copy of this same recognition, and it is now a **no-op on all 203
bundles** — the migration reports 207/207 unchanged without it. Keeping a second,
divergent authority would have meant a re-migration of raw bundles could publish
internal 16-hex machine ids. The anonymizer is now the single authority.

## Tests

- `TestPublicPseudonymFixedPoint` in `tests/unit/core/results/test_anonymization.py`
  — fixed point for result and tuning payloads, emitter/recognizer agreement
  (pins `[:12]` against the accepted width so they cannot drift apart), plus the
  two scoping guards above.
- `test_anonymizing_the_corpus_twice_matches_anonymizing_it_once` in
  `tests/unit/scripts/test_corpus_privacy_invariant.py` — whole-corpus gate,
  guarded against a vacuous scan, in the same module as the existing corpus
  privacy invariant.

### DIFF
--- a/_project/scripts/results_explorer_corpus_migrate.py
+++ b/_project/scripts/results_explorer_corpus_migrate.py
@@ -13,7 +13,6 @@ import argparse
 import hashlib
 import json
 import os
-import re
 import sys
 import tempfile
 from pathlib import Path
@@ -33,7 +32,6 @@ DEFAULT_MANIFEST = BUNDLES_DIR / "path-privacy-migration.manifest.json"
 COMPANION_SUFFIXES = (".plans.json", ".tuning.json", ".applied.json")
 MANIFEST_SUFFIX = ".manifest.json"
 STRUCTURAL_MANIFEST_KEYS = frozenset({"bundle_file", "bundle_hash", "companion_hashes"})
-_PUBLIC_HASH_RE = re.compile(r"^(?:machine|host|path|endpoint|bucket|prefix|arn|table|column|database)_[0-9a-f]{8,64}$")
 
 
 def _sha256(raw: bytes) -> str:
@@ -70,19 +68,8 @@ def _load_json(path: Path) -> tuple[Any, bytes]:
 
 def _public_companion_payload(path: Path, payload: Any, manager: AnonymizationManager) -> Any:
     if path.name.endswith(".tuning.json") and isinstance(payload, dict):
-        return _preserve_public_hashes(payload, manager.anonymize_tuning_payload(payload))
-    return _preserve_public_hashes(payload, manager.anonymize_result_payload(payload))  # type: ignore[arg-type]
-
-
-def _preserve_public_hashes(original: Any, anonymized: Any) -> Any:
-    """Keep already-public identifier hashes stable across repeated migrations."""
-    if isinstance(original, dict) and isinstance(anonymized, dict):
-        return {key: _preserve_public_hashes(original.get(key), value) for key, value in anonymized.items()}
-    if isinstance(original, list) and isinstance(anonymized, list):
-        return [_preserve_public_hashes(old, new) for old, new in zip(original, anonymized, strict=False)]
-    if isinstance(original, str) and _PUBLIC_HASH_RE.fullmatch(original):
-        return original
-    return anonymized
+        return manager.anonymize_tuning_payload(payload)
+    return manager.anonymize_result_payload(payload)  # type: ignore[arg-type]
 
 
 def _manifest_path(bundle_path: Path) -> Path | None:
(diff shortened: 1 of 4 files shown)


joeharris76:
## Correction to this PR's claims (adversarial re-review, 2026-08-04)

Two statements in the description above are wrong. Recording it here rather than
leaving them to be trusted.

**1. "The corpus needs no re-derivation — its stored values were already exactly
one pass from raw."** Verified on `machine_id` only, then generalised. Measuring
the whole payload, a fresh submission and the curated bundle are **identical for
0 of 207 bundles**:

```
environment.platform_runtime.engine_host             195
platform.config/execution.driver_runtime_python_executable  146 each
platform.config.working_dir                          138
platform.raw_config.working_dir                      133
platform.config/raw_config.database_path              42 / 37
platform.config/raw_config.database                   25 / 25
```

Mechanism: the exporter already anonymizes at capture time, so raw bundles
carried `engine_host` as `host_51e3fb6b450f` — already a 12-hex pseudonym — and
#1467 hashed it a **second** time. `machine_id` was 16-hex, not pseudonym-shaped,
so it was hashed exactly once. That is the only reason the spot-check agreed.

So the cross-corpus grouping fix landed for `machine_id` and **not** for
`host_*` / `path_*` / `database_*`. Re-derivation — the original suggestion #3,
dismissed here — is still needed. Tracked as `rederive-corpus-to-single-pass`.

**2. "203 bundles."** The real count is **207**. The 203 came from a flat glob
that missed 4 nested bundles under `results-data/bundles/ssb/**`. The corpus
gate added here uses `rglob` and does cover them, so the gate is sound — only
the quoted figure was wrong.

**What stands:** idempotence is fixed and re-verified (207/207 stable, zero
`find_public_path_leaks` at one and two passes). The prefix- and width-scoped
pass-through, the emitter/recognizer pin, and the removal of
`_preserve_public_hashes` are all unaffected.

Separately, and not caused by this PR: `published-results` — a public branch —
still carries the pre-#1467 bundles, 183/207 of which leak private paths.
Tracked as `published-results-corpus-leak-and-mirror-gap`.

PR #1521, opened 2026-08-04
ci: run the orphaned-commit detector on every develop push

joeharris76:
## The race

`make pr-open` enables squash auto-merge at PR creation. If work continues on the branch afterwards, the PR can satisfy its checks and merge **while a later commit is still being written** — develop gets a partial change, the branch is merged, the PR is closed, and nothing reports that part of the work never landed.

PR #1503 merged carrying only its first commit; the `Results Explorer browser gate` job its own body described never landed. The detector's own header records **three earlier occurrences** in this same remediation.

## Why not fix it in `auto-merge-on-open.yml`

It can't be. That workflow reacts to `synchronize` — and in this race there is no later push to react to *until after the merge has already happened*. Withholding auto-merge was rejected outright: the item's must-preserve requires an ordinary fully-pushed PR to keep merging hands-free.

So the guard is **detection** — and a detector for exactly this class already exists. The gap was cadence, not absence.

## Change

Its develop push trigger was path-filtered to edits of *the detector itself*, leaving the weekly schedule as the only real coverage — up to seven days in which develop is silently missing a change everyone believes landed. Drop the filter so it runs at the moment a squash merge can strand a commit. The schedule stays as a backstop for branches pushed long after their PR closed.

## Verification

Run live against current state, it flags **5 orphans across 3 branches** — including my own `b5c56ea9` from #1503, now allowlisted (content re-landed verbatim via #1504).

4 tests pin: the trigger, the retained schedule, the allowlist escape hatch (one acknowledged orphan must not make every later run red and decay the signal), and — as a must-preserve regression guard — that auto-merge still re-evaluates on `synchronize`, still revokes for soundness paths, and still enables `--auto --squash`. Falsified by restoring the paths filter. Rung 2 (`pytest -k auto_merge`): **36 passed**.

## Heads-up: this will be red on merge

Four orphans remain that are **not mine to resolve**, and two are genuinely lost work:

| Commit | Branch | Status |
|---|---|---|
| `62fa0131` docs(agents): forbid standing agent attribution | `claude/git-identity-cloud-rca-maiixe` | **lost** — `agent-attribution-surfaces.md` absent from develop |
| `670fc009` fix(ci): drop the agent-identity guard that cannot fail | same | **lost** — develop still has the guard |
| `cc01856f` docs(decisions): ADR text | `test/freeze-bulk-path-coverage` | already on develop — allowlist candidate |
| `43fd39fe` merge commit | same | benign |

The detector is telling the truth, and it is non-blocking (push-triggered, not a required check). But develop will show a red detector run until those are re-landed or allowlisted. I have not touched another session's in-flight work — asking before acting.

### DIFF
--- a/.github/workflows/orphaned-commit-detector.yml
+++ b/.github/workflows/orphaned-commit-detector.yml
@@ -18,15 +18,20 @@ on:
     # Weekly, Monday 07:00 UTC (after the submission-validator drift check).
     - cron: "0 7 * * 1"
   workflow_dispatch: {}
-  # Also run when the detector or its allowlist changes, so a regression in the
-  # detector itself (or a bad allowlist edit) surfaces immediately.
+  # Run on EVERY develop push, not just changes to the detector itself.
+  #
+  # The weekly schedule above is a backstop, not a guard: it leaves up to seven
+  # days in which develop is silently missing a change that everyone believes
+  # landed. That is not hypothetical - PR #1503 auto-merged carrying only its
+  # first commit, and the orphan (b5c56ea9) was found by hand days before the
+  # schedule would have fired. A develop push is exactly the moment a squash
+  # merge can strand a commit, so check then.
+  #
+  # The job is read-only and finishes in ~1 minute; running it per merge is the
+  # difference between catching a stranded commit in one build and in one week.
   push:
     branches:
       - develop
-    paths:
-      - "_project/scripts/detect_orphaned_commits.py"
-      - "_project/analysis/known-stranded-commits.txt"
-      - ".github/workflows/orphaned-commit-detector.yml"
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


joeharris76:
Addressed both findings.

**P1 "Run detection after post-merge branch pushes" — correct, fixed in `ef0fde5b`.** You're right that the develop-only trigger is structurally too early: the merge fires the develop push, and the orphan is pushed to the feature branch *afterwards*, so at that moment it does not exist. Dropped the branch filter entirely — feature-branch pushes now catch the orphan as it is created, develop pushes still catch ones stranded by an earlier merge. Falsified by restoring the develop-only filter; `tests/unit/workflows` 131 passed.

**P1 "Reauthor the commit under the resolved human identity" — refuted.** The commit on this branch is:

```
56e45ac046  author=Joe Harris <joeharris76@gmail.com>  committer=Joe Harris <joeharris76@gmail.com>
```

The cited commit `4fbd4b77a101` is not on this branch — `git log origin/develop..origin/fix/pr-open-auto-merge-partial-stack-race-20260804` returns exactly one commit, the one above. The same identity finding appears on #1516, #1517 and #1518; all three are likewise authored and committed by `Joe Harris <joeharris76@gmail.com>`. No reauthoring needed.

PR #1531, opened 2026-08-04
docs(browser-ci): correct the amd64 slowdown arithmetic

joeharris76:
Addresses Codex P2 on #1525 (already merged, so this is a follow-up).

The reviewer is right and the error was mine: **the 10.5-minute figure cannot establish an 11x slowdown, because it includes the ten failures it was being used to explain.**

Each failed data wait deliberately burns its full 10s + 8s + 8s budget plus up to two bounded re-navigations, so the failures alone account for **≥260s of intentional waiting**. Netting that off:

| | |
|---|---|
| Emulated run | 630s |
| Deliberate waiting in the 10 failures | ≥260s |
| Remaining actual work | ≤370s |
| arm64 baseline | 58s |
| Honest ratio | **~6x upper bound**, not 11x |

And even ~6x is an upper bound, since the remaining 370s still contains the six *passing* tests.

Replaced the multiplier with "several times slower, magnitude not established" — all this arm can support.

**The conclusion is unchanged and never depended on the number**: every failure is a data-bound wait exhausting its budget, which any large slowdown produces regardless of cause, and native amd64 is green in CI. The architecture hypothesis stays confounded and unsupported.

Docs-only.

### DIFF
--- a/docs/operations/browser-ci.md
+++ b/docs/operations/browser-ci.md
@@ -225,23 +225,34 @@ with `Rosetta is not installed`) the develop tree fails **10 of 16** WebKit
 the same tree passes 16/16.
 
 That looks like a reproduction and is not one. The emulated run took
-**10.5 minutes** against **58 seconds** on arm64 — roughly **11x slower**. Every
-failure is a data-bound wait exhausting its 10s + 8s + 8s budget, which is what
-an 11x slowdown produces regardless of cause. GitHub's runners are *native*
-amd64, not emulated, and the WebKit lane is green there.
+**10.5 minutes** (630s) against **58 seconds** on arm64.
+
+Do not read that as an 11x slowdown: the 630s *includes the ten failures being
+explained*. Each failed data wait deliberately burns its full 10s + 8s + 8s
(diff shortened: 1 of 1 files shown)


joeharris76:
**P2 "Remove the unsupported 6x upper bound" — you're right, and I'd made the same class of error twice. Fixed in `bff13a3a`.**

The subtraction repairs nothing: the emulated run *terminated ten tests* before they completed their normal workload, while arm64 completed all sixteen. Netting off the budgets leaves work for six completed plus ten aborted tests, which is not comparable to 58s of sixteen completed ones — and polling overlaps application processing anyway.

So I've withdrawn the ratio entirely rather than replacing one unsupported number with another. The section now states plainly that **no ratio can be computed from these two numbers in either direction**, records that both "~11x" and "~6x upper bound" were unsupported and withdrawn, and notes that a defensible figure needs matched successful-phase timings this arm did not collect.

The conclusion never depended on the magnitude and is unchanged: every failure is a data-bound wait exhausting its budget, which any large slowdown produces regardless of cause, and native amd64 is green in CI.

PR #1537, opened 2026-08-04
fix(corpus): re-derive every bundle to a single anonymization pass

joeharris76:
Completes what #1512 claimed but did not deliver. See the
[correction comment](https://github.com/joeharris76/BenchBox/pull/1512#issuecomment-5179528843)
on that PR.

## What was wrong

#1512 concluded the corpus needed no re-derivation because its stored values
were "already exactly one pass from raw." That was verified on `machine_id`
**only** and generalised. It is false for every other identifier family.

The exporter already anonymizes at capture time, so the pre-#1467 bundles
carried `engine_host` as a 12-hex pseudonym *already*, and #1467 hashed it a
second time:

| field | raw (pre-#1467) | curated (develop) | fresh publish | curated publish |
|---|---|---|---|---|
| `engine_host` | `host_51e3fb6b450f` | `host_7e5bc04be41b` | `host_51e3fb6b450f` | `host_7e5bc04be41b` ❌ |
| `machine_id` | `machine_45d1a3ef…` | `machine_3d46427037d2` | `machine_3d46427037d2` | `machine_3d46427037d2` ✓ |

`machine_id` was 16-hex — not pseudonym-shaped — so it was hashed exactly once.
That is the only reason the original spot-check agreed.

Measured across the **whole payload**, fresh submission vs curated bundle:
**identical for 0 of 207**, differing on `engine_host` (195),
`driver_runtime_python_executable` (146 in each of two blocks), `working_dir`
(138 + 133), `database_path` (42 + 37), `database` (25 + 25), and more.

So the same physical machine published under two different pseudonyms depending
on which side it came from — the grouping bug #1512 set out to remove, fixed
only for `machine_id`.

## Change

Restore the pre-#1467 originals and apply exactly one public pass via
`results_explorer_corpus_migrate.py --write`, which since #1512 no longer
carries `_preserve_public_hashes` and so applies a single clean pass. File sets
at the two revisions are identical apart from #1467's own audit manifest, so
nothing is resurrected or dropped.

## Verification

```
fresh submission == curated, whole payload : 207/207 identical   (was 0/207)
private-path leaks over 400 corpus files   : 0
idempotent over 207 primary bundles        : 0 non-idempotent
stored bytes == published bytes            : 207/207
corpus-inventory drift                     : clean (regenerated, 207 bundles)
full fast lane                             : 26755 passed
```

Every `result_id` changes. Still free: `origin/release` carries zero bundles and
`published-results` records no `result_id`.

## New test, and what it deliberately does not do

`test_rederived_corpus_publishes_byte_identically_to_what_is_stored` pins that
the corpus sits **at** the publication fixed point rather than merely converging
to it — publishing rewrites nothing, and a result ID is the same computed from
either side. It fails closed on the realistic regression: re-importing bundles
that never went through the public boundary, which the idempotence gate alone
cannot catch.

It deliberately does **not** verify pass-count. A once-hashed and a twice-hashed
pseudonym are byte-indistinguishable without the pre-anonymization original — I
confirmed the gate reports 0 drift on *both* develop's corpus and this one. That
check is inherently one-time and lives in the commit message with numbers.
#1512's idempotence also makes the double-hash structurally unrepeatable, so
there is nothing left for a recurring test to guard.

## Follow-up

`published-results` will need re-mirroring after this lands so the public branch
picks up the corrected pseudonyms.

Also: the corpus is **207** primary bundles, not the 203 quoted in #1512 — that
figure came from a flat glob that missed 4 nested bundles under
`results-data/bundles/ssb/**`.

### DIFF
--- a/results-data/bundles/amplab_sf001_datafusion_df_20260502_223331_f228f907.json
+++ b/results-data/bundles/amplab_sf001_datafusion_df_20260502_223331_f228f907.json
@@ -27,7 +27,7 @@
     "machine_id": "machine_3d46427037d2",
     "platform_runtime": {
       "collection_status": "partial",
-      "engine_host": "host_7e5bc04be41b",
+      "engine_host": "host_51e3fb6b450f",
       "runtime_type": "dataframe_process",
       "source": "inferred"
     }
@@ -37,7 +37,7 @@
       "type": "zstd"
     },
     "driver_package": "datafusion",
-    "driver_runtime_python_executable": "path_ae50093e082b",
+    "driver_runtime_python_executable": "path_51d816457550",
     "driver_runtime_strategy": "current-process",
     "driver_version_actual": "53.0.0",
     "driver_version_resolved": "53.0.0",
@@ -109,7 +109,7 @@
     "config": {
       "batch_size": 8192,
       "driver_package": "datafusion",
-      "driver_runtime_python_executable": "path_ae50093e082b",
+      "driver_runtime_python_executable": "path_51d816457550",
       "driver_runtime_strategy": "current-process",
       "driver_version_actual": "53.0.0",
       "driver_version_resolved": "53.0.0",
@@ -118,7 +118,7 @@
       "parquet_pushdown": true,
       "repartition_joins": true,
       "target_partitions": 10,
-      "working_dir": "path_4b3c50e1d5f1"
+      "working_dir": "path_446c65ea683f"
     },
     "deployment": {
       "collection_status": "partial",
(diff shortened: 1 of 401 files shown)


joeharris76:
Closing as superseded — the corpus change here is already on `develop`.

The `results-data` tree object is byte-identical on both sides:

```
develop  results-data tree: 09bf77142e68a3f27a42eab44e0183958d320f8a
branch   results-data tree: 09bf77142e68a3f27a42eab44e0183958d320f8a
```

`git diff origin/develop origin/fix/rederive-corpus-single-pass -- results-data` is empty. `develop` reached the same single-pass state independently through #1467 and #1538 (`355cc445`), so this branch's commit `c9414c7` — which does change 400 corpus files relative to *its own* parent — now re-applies a state that is already in place.

The invariant it was establishing is held and enforced on develop: `tests/unit/scripts/test_corpus_privacy_invariant.py::test_rederived_corpus_publishes_byte_identically_to_what_is_stored` passes there, and #1568 pins the corpus to LF so that gate stops being platform-dependent.

Nothing to rebase. Reopen if the tree comparison above is wrong.

PR #1546, opened 2026-08-04
fix(agents): bring the active instruction surface back under budget

joeharris76:
develop has been red since #1538 on both `fast-test` and `lint`: the
agent-instruction audit reports `active instruction bytes 16088 exceed budget
16000`, which fails three tests in
tests/unit/scripts/test_agent_instruction_audit.py and the `agent-instructions`,
`agent-identity`, and `agent-commit-range` guards in `ci-lint`. It also fails
the pre-commit hook, so no commit can land in any worktree until it is fixed.

#1541 added the external-instruction rule to the [COMMIT-IDENTITY-001] block and
trimmed the UAT paragraph to pay for it, but the trim was ~40 bytes short.

The budget is a ratchet (baseline 42258, budget 16000), so this pays the
overage down rather than raising the ceiling. Four redundancy removals, no rule
changes and no anchor changes:

- intro: drop the skill-mirror sentence, which restates the "Skills and
  generated mirrors" section verbatim (-71)
- agent-identity-check: "That is detection, not prevention: it reports drift
  already present" -> "detection, not prevention. It reports existing drift"
  (-15)
- review protocol: drop the trailing "document" after a filename (-9)
- ephemeral clone: "That declaration is ignored" -> "It is ignored" (-16)
- Apple/macOS: "unsuitable for the known multi-service stacks; follow" ->
  "unsuitable for known multi-service stacks; see" (-15)

16088 -> 15972 bytes, 167 -> 165 lines. Every pinned anchor is byte-identical:
the [COMMIT-IDENTITY-001] and [REVIEW-AUTH-001] sections are untouched apart
from the identity-check sentence, which carries no anchor.

Headroom is now 28 bytes. The next change that adds to AGENTS.md, CLAUDE.md,
GEMINI.md, ANTIGRAVITY.md, .claude/commands/pr.md, or
docs/development/agent-review-protocol.md must trim to pay for itself.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,8 +1,7 @@
 # BenchBox Agent Guide
 
 This file is the project authority for agent work. Keep it compact: detailed
-operations belong in the linked docs, and generated skill mirrors belong to
-`/Users/joe/.skill-sync/skills`.
+operations belong in the linked docs.
 
 ## Authority and provenance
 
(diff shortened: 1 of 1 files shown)


joeharris76:
Superseded by #1544, which landed the same unblock (instruction budget 16088 -> 15993) while this was in CI. Closing rather than resolving the conflict: #1544 is already on develop and this branch would only re-trim text it already compressed.

PR #1555, opened 2026-08-04
refactor(core): one source for run modes, phases, and export formats

joeharris76:
Three value sets were literals in two or more places and had already drifted:

- `--mode` accepted {sql, dataframe}; MCP accepted {sql, dataframe, data_only}.
- The seven-phase list was written twice inside the CLI (run.py and
  benchmarks.py) and validated nowhere in MCP.
- Export formats appeared as an inline literal in cli/commands/export.py and
  twice more in mcp/tools/results.py -- with two different lengths, because the
  dispatch tuple and the error message's valid_formats list had diverged.

benchbox/core/constants.py now holds RUN_MODES, VALID_PHASES, QUERY_PHASES,
EXECUTION_TYPES, EXPORT_FORMATS, MCP_RESULT_FORMATS, MCP_MODE_CHOICES, and
MCP_DATA_ONLY_ALIASES; every site above imports from it.

## The data_only ruling

The review framed this as an omission to resolve "either exposed on both
surfaces or entered in the tiered omission ledger". Neither fits, and reading
the code shows why: the two surfaces are not exposing the same kind of thing.

`sql` and `dataframe` are platform CAPABILITIES. The CLI validates `--mode`
against PlatformRegistry.supports_mode(), which answers "can this platform run
queries that way". `data_only` is not a way of running queries -- it means run
none at all. In the CLI it is an EXECUTION TYPE derived from the phase
selection (`_derive_execution_type`, `--phases generate`), and it flows to
ExecutionPipeline._execute_data_only, a different branch entirely from the
sql/dataframe path.

So adding `data_only` to the CLI's `--mode` Choice would not have been a
one-line reconciliation: line 990 of run.py would ask supports_mode() a question
it cannot answer, and resolved_mode would carry an execution type into
capability-shaped branches. It would conflate the two concepts in the CLI to
make a constant look symmetric.

The ruling instead splits them: RUN_MODES is the capability set both surfaces
share, EXECUTION_TYPES holds data_only, and MCP_MODE_CHOICES documents that MCP
accepts the execution type on its `mode` parameter as a shortcut. MCP has a
single `phases` string with no interactive phase selection behind it, so naming
the execution type directly is the only way it can ask for a data-only run. The
asymmetry is real, deliberate, and now documented in docs/reference/mcp.md --
including the observation that it runs the OPPOSITE way from the omission
ledger, which only models CLI controls MCP lacks.

No accepted value was dropped on either surface: `datagen` and `generate`
remain aliases for `data_only`, pinned by a parametrized test.

## MCP phase validation (the gap this closes)

MCP accepted any `phases` string and passed it to
BaseBenchmark.run_with_platform(), so `phases="load,laod"` ran only the load
phase and reported nothing wrong. The CLI has always rejected unknown phases.
validate_phases() now rejects them against VALID_PHASES on BOTH admission paths
-- the synchronous run_benchmark tool and start_benchmark before a durable job
is persisted -- with an error listing the valid phases. A test asserts the
rejection happens before _run_benchmark_impl is called, so a typo cannot get as
far as generating data.

This is the only behavior change: requests that were silently mis-executed are
now rejected. Every previously-valid request still runs.

## Parity tests

Per the item's anti-pattern note, the tests import live objects rather than
grepping source: they read the Click Choice off the registered `run` and
`export` commands, the MCP MODE_CHOICES enum, and the core tuples, and compare
sets. A source grep would rot; these bite on a real divergence.

export.py joins the CLI-surface-drift allowlist because that guard snapshots
decorator SOURCE TEXT -- defining a constant once necessarily rewrites the
decorator, even though the accepted values are unchanged. The rationale points
at the replacement assertion, which is strictly stronger than a source snapshot:
it reads the live Choice values off the command object.

Fast lane 26834 -> 26861 collected.

### DIFF
--- a/benchbox/cli/benchmarks.py
+++ b/benchbox/cli/benchmarks.py
@@ -25,6 +25,7 @@ from benchbox.core.benchmark_registry import (
     get_benchmark_surface,
     validate_scale_factor as core_validate_scale_factor,
 )
+from benchbox.core.constants import VALID_PHASES
 from benchbox.core.schemas import BenchmarkConfig
 from benchbox.utils.printing import quiet_console
 from benchbox.utils.verbosity import VerbositySettings
@@ -919,7 +920,7 @@ def prompt_phases(default_phases: list[str] | None = None) -> list[str]:
         return selected_phases
 
     # Custom selection
-    valid_phases = ["generate", "load", "statistics", "warmup", "power", "throughput", "maintenance"]
+    valid_phases = list(VALID_PHASES)
     unique_phases = _prompt_custom_phases(valid_phases)
 
     console.print(f"[green]✓ Selected phases: {', '.join(unique_phases)}[/green]")
--- a/benchbox/cli/commands/export.py
+++ b/benchbox/cli/commands/export.py
@@ -8,6 +8,7 @@ import click
 
 from benchbox.cli.output import ResultExporter
 from benchbox.cli.shared import console
+from benchbox.core.constants import EXPORT_FORMATS
 from benchbox.core.results.loader import (
     ResultLoadError,
     UnsupportedSchemaError,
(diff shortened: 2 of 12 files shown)


joeharris76:
@codex review

PR #1556, opened 2026-08-04
refactor: close the three residual duplications from the CLI/MCP review

joeharris76:
Three independent sub-moves from `one-engine-residual-dedup`.

## 1. MCP resources and discovery tools share one serializer

benchbox/mcp/resources/registry.py walked the registry a second time to build
benchmark and platform payloads with a slightly different field set.
tools/discovery.py now owns the canonical builders -- build_benchmark_payload
and build_platform_payloads -- and each surface is a PROJECTION of them. Per the
item's anti-pattern note there is no `fields=` switch with per-caller branches;
the projections are separate explicit functions, and a test asserts each surface
delegates to the shared builder rather than rebuilding.

Resource field NAMES are unchanged (the resource keeps its flatter
query_count/query_ids and its estimated_time_minutes; the tool keeps its nested
queries object, schema block, and supports_streams). Three payload VALUES change,
all disclosed:

- **query_ids was empty for every benchmark and is now populated.** The
  resource's private helper probed `bm.query_manager.get_all_queries()` and
  `bm.get_query_ids()`; no benchmark class has either attribute (TPCH exposes
  `get_queries()`), and the helper swallowed the miss into a DEBUG log and
  returned []. Every `benchmark://<name>` resource read has been reporting an
  empty query list. All 21 public benchmarks now return real ids -- tpch 22,
  tpcds 99, tpchavoc 220.
- **The resource platform list is now ordered.** It previously emitted registry
  insertion order; it now inherits the canonical adoption-then-name sort. Same
  51 platforms, same fields, deterministic order.
- **Resource reads now do real work.** Populating query_ids instantiates the
  benchmark and loads its queries: ~0.09s for tpch, ~2.0s for tpcds, against a
  previously instant empty list. That is the cost of the field being correct.

query_count and len(query_ids) disagree for four benchmarks (tpcds_obt 17 vs 89,
read_primitives 136 vs 157, write_primitives 12 vs 112, transaction_primitives
12 vs 23): registry `num_queries` metadata versus the actual loaded query set.
Pre-existing and previously invisible because query_ids was empty. Out of scope
here -- reported rather than silently reconciled.

## 2. Credential metadata moved into the platform registry

setup.py hardcoded a seven-platform table of names, descriptions, and required
credential field names. `required_credentials` now lives in
benchbox/core/platform_registry.py, and setup.py derives its table from
platforms that declare it -- adding a credentialed platform to the registry adds
it to the table. The short display blurbs stay in setup.py: those are
presentation copy, not platform metadata, and keeping them local means the table
reads exactly as it did.

Required-field lists are byte-identical to the old table (asserted). Row order
changes from hand-written to alphabetical, because derived output must be
deterministic and registry insertion order is an implementation detail.
CredentialManager storage behavior is untouched.

## 3. The utils -> cli layering violation is gone

.importlinter carried `benchbox.utils.config_interface -> benchbox.cli.config`
as an ignore entry; its own header says every entry is a real violation that
must be ratcheted down, not grown.

The edge is inverted rather than papered over. utils exposes a registration seam
(set_config_provider / get_config_provider); the adapter class moved to
benchbox/cli/config.py as CLIConfigProvider, and the CLI group callback installs
it at startup. The import now points down (cli -> utils). The ignore entry is
deleted, so lint-imports enforces this forever.

Behavior-neutral in practice: the three `config_manager is None` fallbacks that
reached for the CLI adapter are the only callers, and nothing in production
constructs those classes without passing config_manager explicitly -- every real
caller is a test that supplies its own. Non-CLI callers (MCP, library use) get
SimpleConfigProvider's documented defaults, which is what they already got
whenever benchbox.cli was unimportable.

## Test note

The shared-serializer tests parametrize on `benchmark_id`, not `benchmark`:
pytest-benchmark reserves `benchmark` as a fixture name and raises inside
pytest_runtest_makereport when a parametrized value shadows it, killing the
whole session with an INTERNALERROR instead of failing one test. An earlier
draft hit exactly that. The same draft parametrized all 22 public benchmarks,
which is 22 benchmark instantiations per test; the shared-builder assertion is
structural instead, and only tpch and coffeeshop are built for real.

Fast lane 26834 -> 26846 collected. lint-imports 3/3 contracts kept.


joeharris76:
@codex review

PR #1557, opened 2026-08-04
test: close two Phase 2 review findings

joeharris76:
Adversarial review of the merged Phase 2 diff (#1553-#1556), reviewers prompted
to refute. Two confirmed findings, both in the tests rather than the runtime.

(1) The headline behavior change of #1554 had no test on the production path.

    TestReportThresholdIsNoLongerInert compared a locally-defined lambda against
    is_regression. That proves two expressions in the test file disagree; it
    never called ResultDatabase.detect_regressions. Reverting database.py to the
    superseded filter -- `trend.is_regression and trend.change_pct and
    trend.change_pct > threshold_pct` -- left all 105 tests in the module green,
    so the "--threshold below 10 is silently inert" bug could be reintroduced
    without a single failure.

    The class now drives detect_regressions itself with crafted
    PerformanceTrend records, and covers what the lambda could not: that a row
    selected as a regression does not come back reporting is_regression False,
    and that a None change_pct (the oldest period in a window) is not selected.
    The same mutation now fails two tests.

(2) The utils -> cli dependency inversion in #1556 leaks across tests.

    install_cli_config_provider sets process-global state from the CLI group
    callback, so any test invoking a CLI subcommand makes CLIConfigProvider --
    backed by a real ConfigManager reading ~/.benchbox/config.yaml -- the
    provider for every LATER test in the same xdist worker. A later test
    building ExecutionConfigHelper() without a config_manager then silently
    reads the developer's real config instead of SimpleConfigProvider's
    defaults, and only when it happens to be scheduled after a CLI test on that
    worker. Verified: a fresh process resolves SimpleConfigProvider; after
    `cli run --help` the identical call resolves CLIConfigProvider.

    No production consumer is affected today -- nothing constructs those classes
    without passing config_manager -- so this is a test-isolation defect, not a
    product one. But it is the same shape as the _QUIET leak that produced a
    develop-red on run 28706929881, so it gets the same treatment: an autouse
    post-yield fixture in tests/conftest.py, modelled on and cross-referenced
    with _reset_global_quiet_state, including its module-scope import rationale.

Verified clean in the same pass, with evidence:

- No numeric or classification drift beyond the disclosed detect_regressions
  fix: widening the threshold band by 2x fails 6 tests across the parity and
  exit-code suites.
- Phase validation rejects only genuinely invalid phases. All seven VALID_PHASES
  round-trip; `execute` and `all` are rejected, neither is a documented phase on
  either surface, and the CLI has always rejected them.
- No dangling symbols: the only hits for the four deleted policy/statistics
  helpers are import aliases pointing at their core replacements.
- No weakened MCP security invariants: 90 allowlist/admission/tenancy/ClickHouse
  /Dask/anonymization tests pass.
- No layering regression: lint-imports 3/3 kept, now WITHOUT the
  utils.config_interface -> cli.config ignore entry.
- No further tautologies: mutations that add data_only back to RUN_MODES, drop
  MCP phase validation, and make the resources rebuild their own payload each
  fail 2 tests; restoring the utils -> cli import fails 1.

Fast lane 26987 -> 26991.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -30,6 +30,7 @@ from typing import Any
 
 import pytest
 
+from benchbox.utils.config_interface import set_config_provider
 from benchbox.utils.printing import set_quiet
 
 # Sphinx 11 deprecations in third-party extensions (sphinx_tags, myst_parser, ablog, napoleon).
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1558, opened 2026-08-04
fix(ci): stop the agent-identity guard failing where no identity exists

joeharris76:
**develop-post-merge has been red since #1523**, on:

```
ERROR: unable to resolve Git author identity
ERROR: unable to resolve Git committer identity
make[1]: *** [Makefile:731: agent-identity-check] Error 1
FAILED guards: agent-identity
```

## Cause

#1523 removed the `pr.yml` step that injected a placeholder identity so `make agent-identity-check` would be runnable on an ephemeral runner. That was **correct** — a check fed a known-good identity can never fail.

But `make ci-lint` still calls the same target, and `develop-post-merge.yml` still runs `ci-lint`. There the runner has no git identity at all, so the guard flipped from *always passing* to *always failing*. Both states are equally uninformative.

## Fix

This check exists to reject a **known agent** identity. An absent identity is nothing to judge — reporting it as an error says only "this environment has no git config". Treat unresolvable as not applicable.

Locally the case cannot bite: git refuses to commit without an identity. And `agent-commit-range-check`, which inspects the commits the branch actually carries, remains the merge-time control — unchanged.

## Testing this honestly took three attempts

The absent case **cannot** be reproduced by clearing config: git then synthesises an implicit `user@host` identity. My first two tests passed against the *unfixed* script for exactly that reason — they proved nothing.

The test now drives `_resolved_git_identity` directly, and fails on the unfixed script with the real message:

```
Left contains 2 more items, first extra item: 'unable to resolve Git author identity'
```

A second test pins that the populated agent case still yields two errors, so the skip cannot be widened into a blanket pass. Verified manually: agent identity rejected, human identity accepted, absent skipped.

44 passed.

### DIFF
--- a/_project/scripts/agent_instruction_audit.py
+++ b/_project/scripts/agent_instruction_audit.py
@@ -258,11 +258,26 @@ def audit_git_identity(project: Path) -> list[str]:
     author_name, author_email = identities["author"]
     author_is_human = bool(author_name and author_email) and not _is_agent_identity(author_name, author_email)
 
+    # No resolvable identity is not a violation. This check exists to reject a
+    # *known agent* identity, and an absent one is nothing to judge - so
+    # failing here says only "this environment has no git config", which is the
+    # normal state of an ephemeral CI runner. `make ci-lint` runs this target
+    # and `develop-post-merge.yml` runs ci-lint, so treating absence as an error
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1559, opened 2026-08-04
test: characterize run orchestration before the core run service

joeharris76:
First work unit of `one-engine-core-run-service`, landed on its own because
the item's anti-patterns require characterization BEFORE any orchestration code
moves, and because it surfaced a live defect that should be visible now rather
than discovered mid-extraction.

## The defect this pins

Two implementations derive a benchmark's execution type from its phases:

  benchbox/cli/commands/run.py::_derive_execution_type
  benchbox/mcp/tools/benchmark.py::_map_phases_to_test_execution_type

Compared across all 127 non-empty subsets of VALID_PHASES, they disagree on 48.
The divergence is exactly, and only, the case that selects two of the three
query phases:

  phases                          CLI          MCP
  power,throughput                combined     power
  power,maintenance               combined     power
  throughput,maintenance          combined     throughput

The CLI returns `combined`, which its own comment explains is required so the
adapter executes exactly the requested subset. MCP's if-chain returns whichever
single phase it tests first, so the other requested phase is silently dropped:
an MCP client asking for `phases="power,throughput"` gets a power run and no
indication that throughput never executed.

It went unnoticed because everything else agrees -- all 48 single-query-phase
subsets, all 16 three-query-phase subsets, `load,power` (the documented MCP
default), `generate`, and `load` all match. Only the "pick 2 of 3" case is
wrong, and it is not the common request.

## What this commit does and does not do

It does not fix it. Resolving the divergence is the core run service's job, and
doing it here would be an undisclosed behavior change to the MCP run surface
outside its own item. What this does is make the resolution reviewable: the
current behavior of BOTH surfaces is asserted, so whichever way the run service
unifies them, these expectations must be edited in the same change.

The bound is asserted in both directions -- 48 diverging subsets all involve
exactly two query phases, and all 79 agreeing subsets involve a number other
than two -- so the characterization cannot rot in one direction while looking
healthy in the other.

Also pinned: the agreed derivations that extraction must preserve, order
independence of the derivation, the empty-phase-list case, and that MCP admits
every VALID_PHASES member.

Additive only; no runtime file is touched. Fast lane 26991 -> 27024.

### DIFF
--- a/tests/unit/test_run_orchestration_characterization.py
+++ b/tests/unit/test_run_orchestration_characterization.py
@@ -1,0 +1,169 @@
+"""Characterization of run orchestration, ahead of the core run service.
+
+`one-engine-core-run-service` extracts run orchestration into
+`benchbox.core` below both surfaces. Its first work unit is characterization,
+and its anti-patterns require that extractions move code verbatim and that any
+behavior delta be listed explicitly rather than discovered later.
+
+This module pins what the two surfaces do TODAY, so the extraction can be proven
+behavior-preserving -- or its deltas named. It deliberately does not fix
+anything: the divergence recorded in TestExecutionTypeDivergence is a defect the
+run service is expected to resolve, and pinning it first is what makes that
+resolution reviewable instead of invisible.
+
+Two independent implementations derive a benchmark's execution type from its
+requested phases:
+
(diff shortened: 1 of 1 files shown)


joeharris76:
@codex review

PR #1560, opened 2026-08-04
test: characterize CLI run-config resolution before extracting it

joeharris76:
Work unit w1 of `one-engine-core-run-service`. w2 moves config resolution out
of BenchmarkOrchestrator into benchbox/core/run_service.py, and the item makes
characterization the gate on its must-preserve line "CLI-observable behavior is
unchanged". This pins what _prepare_run_config produces today so the extraction
can be shown to produce the same RunConfig. Additive; no runtime file touched.

Two surface dependencies are what make w2 a move rather than a lift, and both
are pinned here so the extracted signature can take them as resolved inputs
without changing results (core may not import benchbox.cli):

- self.directory_manager, a benchbox.cli.config.DirectoryManager, supplies the
  database path
- self._verbosity supplies the verbose/quiet fields

## Two assumptions the characterization refuted

I wrote this module expecting two behaviors and got neither, which is the
argument for characterizing before moving rather than after:

- **A tuning config does not disambiguate the database path by truthiness.**
  generate_database_filename derives its tuning segment from specific keys --
  `_metadata.configuration_type`, `primary_keys.enabled`, `foreign_keys.enabled`,
  `table_tunings` -- so `unified_tuning_configuration="sorted-keys"` and an
  arbitrary dict both still produce `tpch_sf1_notuning_noconstraints.duckdb`,
  colliding with an untuned run's database file. Only a recognized shape yields
  a different filename. Three tests pin all three cases: bare string collides,
  arbitrary dict collides, `{"_metadata": {"configuration_type": "tuned"}}`
  does not. Had I asserted my assumption, the extraction would have inherited a
  test that passes for the wrong reason.

- **VerbositySettings has no `verbose` constructor argument.** It takes
  (level, verbose_enabled, very_verbose, quiet) and exposes `verbose` as a
  derived attribute. The test now constructs it the way the CLI does.

## What is pinned

Iteration resolution including the `or DEFAULT` fallback that makes 0 and None
behave alike, the max(1, ...) and max(0, ...) clamps, and string coercion; seed
resolution including that 0 is preserved rather than treated as absent
(`is not None`, not truthiness); passthrough of query subset, concurrency,
scale factor, and the three plan-capture flags; the standard execution-type
default; database-path participation of benchmark name and scale factor;
verbosity flow-through; and that resolution is stable across repeats and does
not mutate the caller's options dict -- the property the extraction relies on.

Fast lane 27024 -> 27054 collected, inside the 27550 ceiling.

### DIFF
--- a/tests/unit/cli/test_run_config_resolution_characterization.py
+++ b/tests/unit/cli/test_run_config_resolution_characterization.py
@@ -1,0 +1,255 @@
+"""Characterization of CLI run-config resolution, ahead of the core run service.
+
+`one-engine-core-run-service` w2 moves config resolution out of
+BenchmarkOrchestrator and into `benchbox/core/run_service.py`. Its anti-patterns
+require extraction diffs to read as pure moves, and its must-preserve list makes
+characterization the gate on "CLI-observable behavior is unchanged".
+
+This module pins what `BenchmarkOrchestrator._prepare_run_config` produces today
+across a representative option matrix, so the extracted service can be shown to
+produce the same RunConfig. It touches no runtime file.
+
+Two surface dependencies are what make this a move rather than a lift: the
+method reads `self.directory_manager` (a benchbox.cli.config.DirectoryManager)
+for the database path, and `self._verbosity`. Both are pinned here so the
+extracted signature can take them as resolved inputs without changing results --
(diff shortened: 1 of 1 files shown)


joeharris76:
@codex review

PR #1561, opened 2026-08-04
refactor(core): add the run service with its config-resolution port

joeharris76:
Work unit w2 of `one-engine-core-run-service`: request/plan types and
configuration resolution. No execution orchestration yet -- that is w3.

benchbox/core/run_service.py holds RunRequest, RunPlan, and
resolve_run_config. BenchmarkOrchestrator._prepare_run_config now delegates to
it, so rung 1 (service exists AND the CLI orchestrator consumes it) is
satisfied and the w1 characterization suite exercises the new path.

## Why the signature looks the way it does

Layering is the whole design constraint: core sits below platforms and cli, so
the service may import neither. The two things _prepare_run_config used to
reach for through `self` become parameters:

- `database_path` replaces `self.directory_manager.get_database_path(...)`.
  DirectoryManager is CLI-layer. The CLI computes the path and hands the
  service data; the tuning-aware naming rules stay where they already live in
  benchbox.utils.database_naming, so w1's characterization of that naming --
  including that a bare-string tuning config collides with an untuned run's
  filename -- continues to describe the same code.
- `verbosity` replaces `self._verbosity`, typed as a structural Protocol so the
  CLI's VerbositySettings satisfies it without core depending on the concrete
  type. SilentVerbosity is the value MCP will pass in the adoption item, having
  no console to configure.

## Pure move

The resolution body is moved verbatim. Every derivation is unchanged: the
`or DEFAULT` fallbacks that make 0 and None behave alike, the max(1, ...) and
max(0, ...) clamps, and the `is not None` seed check that preserves a zero
seed. The proof is that w1's characterization suite passes unmodified through
the delegation, plus a direct assertion that the orchestrator and the service
produce identical RunConfig model dumps for the same inputs.

The orchestrator no longer imports the iteration constants; a test asserts that,
so the moved arithmetic cannot quietly reappear next to the delegation.

## Coverage

An AST test asserts run_service.py imports nothing from benchbox.platforms or
benchbox.cli -- the same shape of guard used for the MCP boundary, and stricter
than lint-imports because it names this module specifically. Another calls the
service with SilentVerbosity and no CLI object at all, which is the MCP path
before MCP exists as a caller.

Rungs 1, 2, 3, and 5 pass; rung 4 is presort, which is w5.

### DIFF
--- a/benchbox/cli/orchestrator.py
+++ b/benchbox/cli/orchestrator.py
@@ -21,15 +21,12 @@ from benchbox.core.benchmark_loader import (
 
 # Import from common_types to avoid circular imports
 from benchbox.core.config import BenchmarkConfig, RunConfig
-from benchbox.core.constants import (
-    GENERIC_POWER_DEFAULT_MEASUREMENT_ITERATIONS,
-    GENERIC_POWER_DEFAULT_WARMUP_ITERATIONS,
-)
 from benchbox.core.hooks.platform_hooks import PlatformHookRegistry
 from benchbox.core.platform_config import get_platform_config as _core_get_platform_config
 from benchbox.core.platform_registry import PlatformRegistry
 from benchbox.core.results.driver_metadata import apply_driver_metadata
 from benchbox.core.results.models import BenchmarkResults
+from benchbox.core.run_service import resolve_run_config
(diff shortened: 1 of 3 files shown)


joeharris76:
@codex review

PR #1562, opened 2026-08-04
refactor(core): move presort sort keys into the benchmark registry

joeharris76:
Work unit w5 of `one-engine-core-run-service`.

`benchbox run --presort` carried hardcoded columns in the CLI: `l_shipdate` for
tpch and `ss_sold_date_sk` for tpcds, inside
_resolve_data_organization_payload. Which column a benchmark sorts on is
benchmark knowledge, and a CLI command file is the wrong owner -- adding
presort support to a third benchmark meant editing the CLI.

The columns now live in benchmark_registry.yaml as `presort_table_configs`, and
declaring that key is what makes a benchmark presort-capable. The CLI reads
get_presort_table_configs() and builds its unsupported-benchmark error from
presort_capable_benchmarks(), so that message can no longer name a stale set.

The registry entries are schema-validated, not free-form: catalog_schema gains a
PresortColumn model with a Literal["asc", "desc"] order, so a typo fails at load
time rather than producing a silently unsorted run.

Behavior is unchanged. The resolver's precedence is pinned by tests: an explicit
tuning payload still wins over registry defaults, delta/iceberg presort still
override output_format on a supplied payload, a non-sorted presort value still
passes the tuning payload through, and an unsupported benchmark still raises.
The accessor returns a copy, so a caller cannot mutate the registry through it.

## Rung 4 is broader than its own expectation

Rung 4 runs `! grep -rn 'l_shipdate\|ss_sold_date_sk' benchbox/cli/` and expects
"Presort defaults live in benchmark registry metadata". The expectation is met
-- no presort literal remains in run.py, orchestrator.py, or
execution_pipeline.py -- but the command still exits 1, because those column
names also appear in two places that have nothing to do with presort:

  benchbox/cli/tuning.py:781,786   interactive prompt hint and default for the
                                   tuning wizard's sort-column question
  benchbox/cli/commands/convert.py:166,170   `--partition l_shipdate` in the
                                   command's help-text examples

Neither is a presort default, neither file is in this item's only_modify list,
and deleting a column name from a help example would make the docs worse to
satisfy a grep. I have not contorted them. The rung command over-reaches its
expectation; the expectation itself is satisfied and is covered by a test
asserting run.py carries neither literal.

benchbox/cli/commands/run.py stays at 3130 lines, exactly the module-size
guardrail -- the first draft of this change tripped it by one line, so the
explanatory comment is two lines shorter rather than the limit being raised.

Fast lane 27054 -> 27083.

### DIFF
--- a/benchbox/cli/commands/run.py
+++ b/benchbox/cli/commands/run.py
@@ -65,7 +65,11 @@ from benchbox.cli.tuning_runtime import (
     infer_runtime_tuning_mode,
     resolve_dataframe_tuning_config,
 )
-from benchbox.core.benchmark_registry import get_benchmark_default_scale
+from benchbox.core.benchmark_registry import (
+    get_benchmark_default_scale,
+    get_presort_table_configs,
+    presort_capable_benchmarks,
+)
 from benchbox.core.config import DatabaseConfig
 from benchbox.core.constants import QUERY_PHASES, RUN_MODES, VALID_PHASES
 from benchbox.core.platform_registry import PlatformRegistry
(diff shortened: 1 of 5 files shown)


joeharris76:
@codex review

PR #1563, opened 2026-08-04
ci: fail a PR whose base is not an integration branch

joeharris76:
## The gap

Every workflow with a `pull_request` trigger filters on `develop`, `release` or `published-results`:

| Filter | Workflows |
|---|---|
| `[develop]` | auto-merge-on-open, extension-smoke, gitignore-lint, **pr.yml** |
| `[release, develop]` | docs, results-explorer-browser |
| `[release]` | lint, perf-smoke, test, validate-release-pr |
| `[published-results]` | validate-submission |

A PR based on another feature branch matches **none** of them, so it triggers nothing at all — no `pr.yml`, no `ci-required-result`, no browser lane. The PR page looks calm rather than alarming: there are no failing checks because there are no checks. The content reaches develop only when the parent merges, at which point it has never been validated on its own and is attributed to the parent's PR.

**Two are live right now:** #1520 based on `claude/mcp-v2-modin-engine`, #1515 on `claude/mcp-v2-duckdb-threads`.

I found this while auditing which PRs the browser-gate ruleset flip would strand — I had assumed their missing gate check meant they predated the workflow, when in fact they receive no CI at all.

## Detect, not prevent

A `make pr-open` refusal would only cover PRs opened through that path; the web UI and `gh pr create` bypass it. A workflow with no branch filter is authoritative for every PR however it was created. Prevention in `pr-open` remains a possible addition, not a substitute.

## Why it can't repeat the failure it fixes

The guard carries **no** `branches` and **no** `paths` filter, so it always runs and therefore always reports — it cannot become the never-reporting required context it exists to prevent. The filter is absent rather than set to a wide list because a list goes stale as branches are added and absence does not. `edited` is in `types` so retargeting an open PR re-evaluates instead of keeping a stale green.

## Tests

Five, including one that pins the **premise** — that every other PR workflow still filters to integration branches — so this guard becoming redundant is noticed deliberately rather than discovered later.

Falsified two ways:
- deleting the workflow → 4 of 5 fail
- adding `branches: [develop]` → `pr-base-guard filters on branches, so the stacked PRs it exists to catch would not trigger it`

`tests/unit/workflows`: **140 passed**.

## Note for the two live PRs

Once this lands, #1520 and #1515 will show a red base-guard check. That is correct and is the point — but it means they need retargeting at develop or folding into their parents before they can be treated as validated.

Closes `stacked-prs-receive-zero-ci-20260804`.

### DIFF
--- a/.github/workflows/pr-base-guard.yml
+++ b/.github/workflows/pr-base-guard.yml
@@ -1,0 +1,55 @@
+name: PR base guard
+
+# Every other PR workflow filters on `branches: [develop]` (or release /
+# published-results), so a PR whose base is another feature branch triggers
+# NOTHING - not pr.yml, not ci-required-result, not the browser lane. The PR
+# page then looks calm rather than alarming: there are no failing checks
+# because there are no checks at all, and the content reaches develop later
+# attributed to the parent's PR, having never been validated on its own.
+#
+# This workflow carries no `branches:` filter, so it is the one check such a
+# PR does get. On a normal base it passes in seconds; on a stacked base it
+# fails loudly and says what to do.
+#
+# It must never become a required context that cannot report - the failure it
+# exists to fix. That is why the filter is absent rather than set to a wide
+# list: a list can go stale as branches are added, absence cannot.
+
+on:
+  pull_request:
+    # `edited` is included so retargeting an existing PR re-evaluates. Without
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1564, opened 2026-08-04
refactor(core): move run execution into the service behind an adapter factory

joeharris76:
Work unit w3 of `one-engine-core-run-service`: execution orchestration ports
into benchbox/core/run_service.py, with adapter construction injected by the
surface.

## The seam

Building an adapter means importing benchbox.platforms, which core sits below,
so execute_run takes an AdapterFactory Protocol instead. The CLI passes a
closure over its console, verbosity, platform config, and benchmark instance;
MCP will pass its own in the adoption item. Returning None is meaningful -- a
data-only run has no database to adapt -- and that is covered.

Everything else the lifecycle needs was already core: LifecyclePhases,
ValidationOptions, and run_benchmark_lifecycle all live in
benchbox.core.runner.runner, PlatformRegistry and is_dataframe_execution in
core, apply_driver_metadata in core.results. The adapter really was the only
platforms-shaped dependency, which is why one injection point is enough.

## What moved, and what deliberately did not

Moved: the requested-phases stamp that combined mode reads downstream, the
phase-list to LifecyclePhases mapping, ValidationOptions construction,
execution-mode detection (explicit mode wins, else registry default upgraded to
dataframe), the run_benchmark_lifecycle call, and the driver-metadata
enrichment.

Stayed in the CLI, per the item's interaction-scoped anti-pattern: all console
output, the credential-setup retry (which recurses into execute_benchmark after
an interactive prompt), and the cloud-platform "executing without load phase"
warning. That warning was inline in the middle of the moved block, so it is now
_warn_on_execute_without_load -- same console string, same trigger conditions,
called before the port. Tests assert core defines no credential function, makes
no print call, and imports nothing from rich or benchbox.utils.printing, and
that the CLI still owns both the warning and the retry.

## Test seam relocation

Six test modules patched `benchbox.cli.orchestrator.run_benchmark_lifecycle`.
The call now lives in core, so those patches are repointed to
`benchbox.core.run_service.run_benchmark_lifecycle`. The assertions are
unchanged -- same kwargs, same behavior -- only the patch target moved, which is
what a pure move should require.

Rungs 1, 2, 3, and 5 pass. Fast lane 27066 -> 27085.

### DIFF
--- a/benchbox/cli/orchestrator.py
+++ b/benchbox/cli/orchestrator.py
@@ -24,15 +24,8 @@ from benchbox.core.config import BenchmarkConfig, RunConfig
 from benchbox.core.hooks.platform_hooks import PlatformHookRegistry
 from benchbox.core.platform_config import get_platform_config as _core_get_platform_config
 from benchbox.core.platform_registry import PlatformRegistry
-from benchbox.core.results.driver_metadata import apply_driver_metadata
 from benchbox.core.results.models import BenchmarkResults
-from benchbox.core.run_service import resolve_run_config
-from benchbox.core.runner.dataframe_runner import is_dataframe_execution
-from benchbox.core.runner.runner import (
-    LifecyclePhases,
-    ValidationOptions,
-    run_benchmark_lifecycle,
-)
+from benchbox.core.run_service import execute_run, resolve_lifecycle_phases, resolve_run_config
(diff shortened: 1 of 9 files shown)


joeharris76:
@codex review

PR #1565, opened 2026-08-05
fix(agents): name the audit check that failed and give the budget headroom

joeharris76:
Closes the misreporting half of `agent-identity-hook-misreports-failing-check`.

## The defect

The pre-commit hook runs the **whole** instruction audit through one entry
point (`make agent-identity-check`) but was named *"reject stale agent Git
identity"*. Pre-commit prints the hook **name**, not the command, when a hook
fails, so any of ~20 unrelated failures — byte budget, policy anchors,
scenarios, forbidden patterns — surfaced as a Git config problem. A byte-budget
failure on develop cost real diagnosis time before the true cause was found.

## Change

| | |
|---|---|
| Attribution | every error is tagged `budget:`, `surface:`, `review-policy:`, `commit-policy:`, `scenarios:`, `git-identity:`, `commit-range:` |
| Output | `agent-instructions: FAIL (budget)` + a `failing_checks` field in `--json` |
| Hook | keeps its id, renamed to *"agent instruction + Git identity audit"* |
| Headroom | a metric at ≥97% of its ceiling now warns, with the remaining margin |

Existing tests assert error **substrings**, not equality, so the prefix does not
break them.

## Why the headroom matters right now

Develop was independently returned to green while this branch was in progress
by tightening AGENTS.md wording. That lands active bytes at **15993 against a
16000 ceiling — 7 bytes of room**. The next one-line addition to any of the six
active files reddens develop again for everyone, and under the current output it
would once more be reported as a stale Git identity.

So the `Audit evidence provenance` section moves out of the always-loaded
surface into its own doc. It is a records convention for `_project/audits/*.md`,
enforced mechanically by `audit_sha_check.py` — not review behaviour every
session must carry, and AGENTS.md already says detailed operations belong in
linked docs.

```
active bytes  15993 -> 14894   (budget 16000, unchanged)
headroom          7 -> 1106
```

No policy ID, anchor, or scenario text was touched; no doc links the moved
section by anchor.

## Deliberately not done

- **The budget is not raised.** The item's must-preserve forbids papering over
  content growth, and the budget is the forcing function.
- **The headroom band is a warning, never an error.** A second failing threshold
  would just be a lower budget. `main()` already excludes warnings from exit
  status.

## Verification

```
make agent-identity-check                     exit 0   (was exit 2 at this branch's base)
pytest tests/unit/scripts/test_agent_instruction_audit.py   50 passed
ruff check / ruff format                      clean on both changed Python files
forced over-budget corpus                     "FAIL (budget)" + "ERROR: budget: ..."
```

Five new tests: error attribution, a budget failure attributed to `budget` and
not identity, `failing_checks` ordering, headroom firing near the ceiling and
staying silent with room to spare, and the hook name not claiming a single check.

## Correction to the tracker record

`w0`'s evidence was written against this branch's original base (`916d9533ab`),
where the audit failed at 16088 bytes. Develop was fixed independently before
this landed, so this PR's contribution is headroom and attribution, not the
initial restoration to green. The measurements above are against current
develop.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -133,8 +133,12 @@ repos:
         stages: [pre-commit]
         files: ^(AGENTS\.md|CLAUDE\.md|GEMINI\.md|ANTIGRAVITY\.md|\.claude/(commands/|settings\.json)|\.claude/skills/|docs/(development/agent-|operations/agent-instruction)|_project/(evals/agent-instructions/|scripts/agent_instruction_audit\.py)|skill-sync\.(yaml|lock))
 
+      # Runs the whole instruction audit, not only the identity check, so the
+      # name cannot claim one check. The script prints `FAIL (<checks>)` naming
+      # what actually failed - a byte-budget failure used to surface here as a
+      # Git identity problem.
       - id: agent-git-identity
-        name: reject stale agent Git identity
+        name: agent instruction + Git identity audit
         entry: make agent-identity-check
         language: system
         pass_filenames: false
--- a/_project/scripts/agent_instruction_audit.py
+++ b/_project/scripts/agent_instruction_audit.py
@@ -8,6 +8,7 @@ import json
 import os
 import re
 import subprocess
+from collections.abc import Iterable
 from dataclasses import asdict, dataclass
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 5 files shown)


joeharris76:
@codex review

PR #1567, opened 2026-08-05
fix: stop arming auto-merge at PR creation

joeharris76:
## Why

Arming auto-merge at PR creation is only correct if nothing more will be pushed — and the usual reason something more *is* pushed is **review feedback**, which arrives after the PR exists. A PR armed at creation can satisfy its checks and merge while the follow-up commit is still being written, leaving develop with a partial change and the remainder orphaned on a closed branch.

**Three times in one session:**

| PR | What was stranded |
|---|---|
| #1503 | the `Results Explorer browser gate` job its own PR body described |
| #1521 | the commit fixing its own Codex review finding |
| #1531 | the commit fixing its own Codex review finding |

No local check can prevent this — in all three the working tree was **clean** when the PR was opened. And `auto-merge-on-open.yml` reacts to `synchronize`, which is too late when the merge has already happened. So the only real prevention is to stop arming at creation.

## Change

- `make pr-open` opens the PR and **withholds** auto-merge, printing how to arm.
- `make pr-ready` arms it — the branch is declared final.
- `make pr-open READY=1` opens and arms in one step, so a branch you already know is finished still merges hands-free.

Arming lives in **one** place, `pr-arm-auto-merge`, used by both entry points, so the soundness withhold cannot drift between copies. `gh pr merge --auto --squash` now appears **exactly once** in the Makefile.

## Must-preserves

- **Hands-free retained** — pinned by a test asserting both `pr-ready` and the `READY=1` path exist, so withholding can't become a dead end.
- **Soundness withholding retained** — a test asserts the soundness predicate is consulted *before* arming, so moving the enablement point can't quietly drop the control that keeps oracle-adjacent changes off the hands-free path.

## Verification

Falsified against the pre-change Makefile — **3 of 4** new tests fail, including:

```
AssertionError: pr-open arms auto-merge directly, so a PR can merge before a later commit is pushed
AssertionError: no pr-ready target; withholding would be a dead end
```

`tests/unit -k auto_merge`: **47 passed**. `tests/unit/workflows`: **138 passed**.

Both runbooks updated so the documented flow matches — this changes the daily loop, so the docs had to move with it.

## Tradeoff, stated plainly

For a finished PR this is one extra command. That is a real if small cost, accepted deliberately: the alternative left three recovery PRs in a single session, two of them recovering review fixes that were themselves stranded.

Closes `pr-open-auto-merge-prevent-not-just-detect-20260804`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ POOL_CLAIM_MARKER_STALE_SECONDS ?= 600
 # truth instead of repeating the four-deep nested expansion.
 POOL_REPO_CMD = basename "$$(dirname "$$(realpath "$$(git rev-parse --git-common-dir)")")"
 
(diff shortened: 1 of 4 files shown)


joeharris76:
@codex review

PR #1572, opened 2026-08-05
fix: gate ci-lint guards that are meaningless on a CI runner

joeharris76:
## Problem

`develop-post-merge.yml` runs `make ci-lint` on an ephemeral GitHub-hosted runner as a blocking gate wired into auto-revert. Two of its ~24 guards read state that only exists on a developer machine, and each failed a different way:

- **`agent-identity-check`** resolves `git config user.*`, which a runner does not have. Before #1558 that reddened post-merge runs for an environment fact. After #1558 it passes — but only because an absent identity now returns `[]`, and because the recipe injected a synthetic `BenchBox CI` identity. Either way it *cannot fail there*, so it verified nothing.
- **`skill-sync-check`** shells out to `$(SKILL_SYNC)` — `/Users/joe/Developer/skill-sync/dist/cli/index.js`, a hardcoded developer path. On a runner it hits the target's own "not installed; skipping" branch and exits 0. **A silent no-op that reads as coverage in the Actions log.**

The second mode is the dangerous one: nobody investigates a green check.

The inline `if [ "$GITHUB_ACTIONS" = "true" ]` block that fed the synthetic identity was itself the smell — a one-off patch for one member, with no general rule behind it.

## Fix

One declarative table (`RUNNER_INAPPLICABLE_GUARDS` in `_project/scripts/ci_lint_environment_gate.py`), consulted uniformly. A guard not listed always runs, on a runner exactly as locally. Adding a third exception is a table entry, not another copy-pasted conditional.

**Exactly 2 of ~24 guards are gated.** `agent-commit-range-check` — the real merge-time authorship control — is never gated and is asserted so in three separate tests.

## Review found the gate reproducing its own defect

The first implementation used the gate process as the `if` condition: exit 0 to run, exit 1 to skip. Every way of failing to *reach* a decision also exits non-zero — `uv` absent, broken venv, script renamed, any traceback — so a broken gate was indistinguishable from a deliberate skip and the guard vanished with nothing in `failed`. Demonstrated:

```
$ if uv run -- python .../ci_lint_environment_gate_TYPO.py agent-identity
-> guard SILENTLY SKIPPED
$ if PATH=/usr/bin:/bin uv run -- python .../ci_lint_environment_gate.py agent-identity
-> guard SILENTLY SKIPPED
```

The decision now travels on stdout as a single token; the recipe skips only on an exact `SKIP` and anything else runs the guard. The reason moves to stderr so it still reaches the log.

Review also removed the synthetic `GIT_CONFIG` identity, now dead. `tests/system/test_ci_lint_parity.py` pinned that exact text, so leaving both would have shipped a **newly vacuous test inside the change whose purpose is removing vacuous passes**. That test now pins the gating and asserts the synthetic identity has not come back. Item scope was widened to `tests/system/**` with the reason recorded on the tracker.

One test asserted `main() == 0` to mean "run"; under the new contract `main` always returns 0, so it had gone vacuous. It now asserts the emitted decision.

## Falsification

Three independent neuters, each caught by a disjoint subset:

| Neuter | Caught by |
|---|---|
| Revert Makefile wiring, keep the module importable | 3 wiring tests, `AssertionError` not `ImportError` |
| Keep wiring, make `runs_on_runner` always return `True` | 3 classification/CLI tests |
| Revert to the exit-status construct | `TestGateFailsClosed` — takes the recipe's own lines, swaps the gate for an uninvokable command, runs them; shell reports `SKIPPED` where `RAN` is required |

The third fails *behaviourally* rather than on a regex miss — an earlier version of it failed with "could not locate the gate condition", which would not have survived a plausible rewrite.

## Coverage

Two guards now run in **zero** places on CI, both previously vacuous there:

- `agent-identity` — no environment anywhere gives a runner a real developer identity. `pr.yml`'s `code-lint` job already omits it for the same reason.
- `skill-sync-check` — the real check is `pr.yml`'s `guard-skill-sync-verify`, a pinned-SHA network checkout gating every PR. Unchanged.

Everything else, including `agent-commit-range-check`, is untouched and unconditional.

## Verification

`test_ci_lint_environment_boundary.py` 16 passed · `test_agent_instruction_audit.py` 50 passed · `test_ci_lint_parity.py` 8 passed · `tests/unit/scripts` 1516 passed · `make -n ci-lint` parses.

## Noted, not fixed

`uv-lock-revision-check` is a no-op whenever `ci-lint` runs on a clean checkout with no explicit `BASE_REF` — `HEAD:uv.lock` always equals the working tree there. That is not a *developer-machine* assumption (it is equally a no-op locally on a clean tree), so it is out of this defect's scope. Flagged rather than silently left.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -903,14 +903,19 @@ guards-fix:
 # `if $GITHUB_ACTIONS` -- see its module docstring); a guard not listed there
 # always runs, on a runner exactly as it does locally.
 #
-# The `GIT_CONFIG_COUNT=2 ... / unset ...` pair still wrapping
-# agent-identity-check below is now dead on a runner: the gate above skips
-# the guard between them, so nothing ever consumes the synthetic identity it
-# exports. It is kept verbatim rather than deleted because
-# tests/system/test_ci_lint_parity.py::test_ci_lint_identity_check_supplies_runner_identity
-# pins this exact recipe text; deleting it needs that test updated in the
-# same change, which is out of scope here (tests/system/** is not part of
-# this fix's authorized paths).
+# The gate reports its decision on stdout and the recipe skips only on the
+# exact token `SKIP`. Deliberately NOT `if <gate-command>; then`: every way of
+# failing to reach a decision (uv absent, broken venv, the script renamed, a
+# traceback) also exits non-zero, so using the exit status would make a broken
+# gate indistinguishable from a deliberate skip and silently drop the guard
+# with nothing recorded in `failed`. Anything other than `SKIP` runs the guard.
+#
+# The synthetic `GIT_CONFIG_*` identity that used to wrap agent-identity-check
+# here was removed with the gate: the guard no longer runs on a runner, so
+# nothing consumed it. Leaving it would have left
+# tests/system/test_ci_lint_parity.py asserting recipe text with no effect --
+# a vacuous pass inside the change whose whole purpose is removing vacuous
+# passes. That test now pins the gating instead.
 ci-lint:
 	@echo "Running CI lint checks..."
 	@case " $(MAKEFLAGS) " in *" n "*|*" -n "*|*" --just-print "*) echo "Dry-run: ci-lint guards suppressed"; exit 0;; esac; \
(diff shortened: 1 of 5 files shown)


joeharris76:
@codex review

PR #1573, opened 2026-08-05
fix(ci): report a real check on published-results mirror PRs

joeharris76:
## Problem

[#1542](https://github.com/joeharris76/BenchBox/pull/1542) has an **empty `statusCheckRollup`** — zero checks, not zero failures.

The cause is not a missing workflow. `validate-submission.yml` lives on `published-results`, triggers on `pull_request` → `published-results` with `paths: results-data/bundles/**`, and **all 100 of #1542's changed files are under that path**. It should have run.

It didn't because the PR's author is `app/github-actions`: `sync-results-data-to-published.yml:195` opens it with `secrets.GITHUB_TOKEN`, and GitHub never starts `pull_request`-triggered runs for `GITHUB_TOKEN` events — documented recursion prevention, not something a trigger filter can reach. The irony is that `validate-submission.yml`'s own comment already documents that this mirror PR is opened with `GITHUB_TOKEN`; nobody joined that to "so this workflow never runs on it."

## Fix — token-free

The sync workflow validates the exact content it is about to propose (the same two gates `validate-submission.yml` would run: `validate_submission.py` and `generate_corpus_inventory.py --check`) and posts the verdict as a `corpus-mirror/validate-bundles` commit status on the mirror branch head.

A commit status is a Statuses-API write, not a workflow trigger, so it is unaffected by the recursion rule and needs **no new credential** — only `statuses: write` on the token the workflow already holds.

## Scope: what's meaningful for a corpus-only branch

`published-results` carries `results-data/`, a slim validation subset of `benchbox/`, two scripts, and docs. Develop's lint/type/pytest matrix has nothing to check there, and the ADR already rejected routing corpus contributions through it. So the mirror PR is judged by exactly the bar a contributor submission PR is judged by — reusing the identical vendored scripts. A test pins that no develop-matrix command creeps into the sync workflow.

## Review found an availability regression

The status and PR-opening steps sit *after* the new setup and validation steps. GitHub skips every later step once one fails, so a flaky `setup-uv` — or any error the validate step's `set -uo pipefail` capture doesn't convert into a reported state — would skip **both**: no status *and* no mirror PR.

That is strictly worse than the blind PR being fixed. Before these steps existed the PR was always created; after, develop's Corpus Drift Check could stay red with nothing open to explain why.

Fixed: both steps run under `always()`, an absent verdict posts as `error` rather than defaulting, and the failure gate is `!= 'success'` rather than `== 'failure'`. **Never default an absent verdict to success** — a green check on content nothing examined is worse than an empty rollup, because it stops anyone looking.

## What this does NOT fix

`published-results` carries exactly one workflow file, so `pr-base-guard.yml` cannot run there — a guard only runs where its file exists. And it cannot be put there by automation: **`GITHUB_TOKEN` cannot push changes under `.github/workflows/` regardless of the `permissions:` block**, which the ADR already documents for `validate-submission.yml` edits. Porting it is a manual maintainer step, written up in the runbook as tracked debt rather than left silent.

Two limits are recorded so a green badge isn't over-read:

- **Advisory, not enforcing.** No ruleset targets `published-results`, so a red mirror PR can still be merged. The draft state and maintainer review are what enforce.
- **Self-attested.** The run that builds the content validates it, so it catches bad content but not a later manual push to the mirror branch. A `--force-with-lease` push from another run re-posts against the new SHA, so a stale green cannot ride on new bytes.

## Falsification

| Neuter | Result |
|---|---|
| Validate step rubber-stamps `STATE="success"` without running either validator (name/id/outputs unchanged) | 1 failure — `validate step never invokes scripts/validate_submission.py` |
| Revert the `always()` guards, the `error` default, and the failure gate | exactly the 3 new availability tests fail; the other 6 pass |

`tests/unit/workflows` 153 passed · `test_stacked_pr_base_guard.py` 5 passed · `pr-base-guard.yml` untouched, still no branch or path filter.

### DIFF
--- a/.github/workflows/sync-results-data-to-published.yml
+++ b/.github/workflows/sync-results-data-to-published.yml
@@ -282,12 +282,26 @@ jobs:
         # SHA, is what makes the PR's statusCheckRollup non-empty. It must carry
         # the same condition as "Open or update draft PR" so every mirror PR gets
         # it, not just some.
-        if: steps.drift.outputs.has_drift == 'true' && steps.build.outputs.has_changes == 'true'
+        #
+        # `always()`: without it, a step-level failure anywhere above (a flaky
+        # `astral-sh/setup-uv`, a `git diff` error inside the validate step —
+        # anything the validate step's own `set -uo pipefail` capture does not
+        # convert into a reported state) skips every later step, so no status
+        # AND no mirror PR. That would be a new way for the corpus mirror to
+        # silently not exist, which is worse than the blind PR this change is
+        # removing: develop's Corpus Drift Check would stay red with nothing
+        # open to explain it. An unreachable verdict is reported as `error`.
+        if: always() && steps.drift.outputs.has_drift == 'true' && steps.build.outputs.has_changes == 'true'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           STATE: ${{ steps.validate.outputs.state }}
           DESCRIPTION: ${{ steps.validate.outputs.description }}
         run: |
+          # Empty means the validate step never reached its `GITHUB_OUTPUT`
+          # write. Never default this to `success`: an absent verdict is the
+          # one case where the PR carries no evidence either way.
+          STATE="${STATE:-error}"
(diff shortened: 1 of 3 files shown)


joeharris76:
@codex review

PR #1574, opened 2026-08-05
fix(core): gate transactional staging reuse on a provenance manifest

joeharris76:
## Problem

`is_setup()` in `transaction_primitives` and `write_primitives` decided staging
readiness with a bare `SELECT COUNT(*)`. Staging tables left over from a
scale-0.01 run therefore satisfied `is_setup()` for a scale-1.0 run against the
same physical database: `setup()` was skipped, the benchmark executed against the
wrong data volume, and the run reported a PASS with meaningless timings. That is
a silent wrong-results bug, not a hygiene issue.

Closes TODO `transactional-staging-reuse-ignores-provenance-20260805`.

## Change

Two commits:

1. **`fix(core)`** — `setup()` now writes a single-row-per-benchmark
   `benchbox_staging_manifest` table (benchmark, scale, spec_version,
   source_digest, created_at) through a shared helper on
   `TransactionalBenchmarkBase`, so the two families cannot drift the way the
   plain-COUNT checks did. Both `is_setup()` overrides delegate to a shared
   `_staging_manifest_matches()` that requires an exact match *in addition to*
   the existing populated-table check.

   A missing manifest — true of every pre-existing database, since no prior code
   ever wrote one — is deliberately **not** treated as "probably fine". It
   returns `False` and self-heals with one rebuild on next use. Defaulting it to
   reuse-eligible would just reinstate the bug.

2. **`test(core)`** — updates the one out-of-scope test that asserted the old
   contract. `test_transaction_primitives_reset_and_setup_checks` asserted
   `is_setup()` was `True` for a connection whose staging tables were populated
   directly by the `_StatefulConn` constructor — no `setup()` call, and a
   `reset()` with `_populate_staging_table` stubbed out, so no manifest is ever
   written. That is precisely the legacy populated-but-unmanifested database the
   first commit makes ineligible for reuse, so the assertion inverted.

## Why the test asserts the new behavior instead of faking a manifest

The alternative was teaching `_StatefulConn` to recognise the manifest INSERT so
the original `is_setup() is True` assertion could stand. That mock matches
`SELECT COUNT(*)` on table name alone and discards the `WHERE` clause, so a
manifest-aware version would report a match for *any*
benchmark/scale/spec/digest — the assertion would pass even against a broken
match predicate, proving nothing about the gate being added.

So the edge-case test now pins the mechanism rather than the boolean: `reset()`
issues no manifest INSERT, `is_setup()` is `False`, and a manifest query *does*
appear in `conn.executed` — showing the staging-count loop passed and the new
gate is what rejected it. The `conn_zero` case additionally asserts it
short-circuits before the manifest is consulted, keeping the two `False`
assertions on genuinely distinct branches. It references
`_STAGING_MANIFEST_TABLE` rather than a hardcoded name, so a rename fails loudly
instead of going vacuous.

Real manifest round-trips run against live DuckDB in the new
`tests/unit/core/transactional/test_staging_provenance.py`: matching manifest
skips rebuild, scale mismatch forces rebuild, legacy unmanifested database forces
rebuild then self-heals, and digest/spec-version sensitivity.

`_StatefulConn` and `test_write_primitives_setup_walk` are unchanged.

## Verification

- `make pr-preflight` — all lint, artifact-hygiene, DDL-drift, oracle-coverage,
  dependency-audit, and spellcheck gates pass.
- Fast suite: **27138 passed, 19 skipped, 0 failed**. Run single-threaded on the
  documented lock-free path because another worktree held the shared test lock;
  the earlier blocked stage was lock contention, not a failure.
- The updated assertion was confirmed to **fail** against the pre-fix
  `is_setup()` (`assert True is False`), so it is a real regression guard rather
  than a test that merely tolerates the new code.
- Not a soundness path per `_project/scripts/auto_merge_soundness_paths.py`.
  Auto-merge intentionally not armed.

### DIFF
--- a/benchbox/core/transaction_primitives/benchmark.py
+++ b/benchbox/core/transaction_primitives/benchmark.py
@@ -465,6 +465,12 @@ class TransactionPrimitivesBenchmark(TransactionalBenchmarkBase["OperationResult
                     except Exception:
                         status[table_name] = 0
 
+            # Record this setup()'s provenance (benchmark, scale, spec version,
+            # source digest) so is_setup() can require an exact match rather
+            # than trusting a bare COUNT(*) that a stale staging set from a
+            # different scale/seed would also satisfy.
+            self._write_staging_manifest(connection, required_tables)
+
             self.log_verbose(f"Setup complete: {status}")
 
             return {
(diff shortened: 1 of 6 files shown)


joeharris76:
@codex review

PR #1577, opened 2026-08-05
fix/unvalidated submit terminal state

joeharris76:
- **fix(results): classify unvalidated runs distinctly from schema violations**
- **test(uat): pin unvalidated cells stay PASSED through run_cell**


joeharris76:
@codex review

PR #1584, opened 2026-08-05
fix(ci): make corpus drift canary directional

joeharris76:
Classify develop-ahead vs published-only so a scheduled canary never
recommends a full mirror that would delete community submissions. Add a
published-tip privacy scan and unit coverage for both paths.


joeharris76:
@codex review

PR #1585, opened 2026-08-05
docs(privacy): close retained-field salt decision

joeharris76:
Record keep-empty OSS default with residual confirmation oracle on
retained published identifiers; reject baked-in salts and one-time
rehash. Extend recovery-gate tests for residual policy and operator salt.


joeharris76:
@codex review

PR #1586, opened 2026-08-05
docs(adr): re-examine published-results history retention for oracle

joeharris76:
Record that empty-salt pseudonym recoverability was weighed after the
2026-08-04 decision; retention still holds because rewrite does not
retract already-public content.


joeharris76:
@codex review

PR #1587, opened 2026-08-05
fix(results): omit empty client_host after public identifier drop

joeharris76:
When client_host only carried machine_id, the drop left a hollow {}.
Omit those empty optional maps; keep non-empty host profiles intact.


joeharris76:
@codex review

PR #1588, opened 2026-08-05
test(submit): pin uncertain+translation-fallback contract

joeharris76:
Classify uncertain with translation fallback as unvalidated (not a UAT
cell failure); translation fallback alone remains schema_violation.


joeharris76:
@codex review

PR #1590, opened 2026-08-05
fix(results): drop alias spellings of unread identifier fields

joeharris76:
Omit workdir/datadir/pythonexecutable-style aliases at the public
boundary so empty-salt path tokens cannot reappear under alternate keys.


joeharris76:
@codex review

PR #1591, opened 2026-08-05
fix(results): split submit schema_violation from bundle load errors

joeharris76:
Path-level unreadable/unparseable artifacts become bundle_load_error.
schema_violation remains for loaded results that fail clean-pass integrity.


joeharris76:
@codex review

PR #1595, opened 2026-08-05
fix(mcp): scrub dsn/connection_string/private_key/sas/pat assignments

joeharris76:
## Summary

Extend MCP exception-text scrubbing so remaining credential assignment forms no longer leak through `make_execution_error` / `scrub_secret_material`:

- `dsn=`
- `connection_string=` / `connection-string=`
- `private_key=` / `private-key=`
- `sas=`
- `pat=` (exact key only — must not match `path=`)

### Approach

`_SECRET_KEY_PATTERN` now includes the expandable keys (`dsn`, `connection[_-]?string`, `private[_-]?key`, `sas`) alongside the existing password/token vocabulary, and treats `pat` as an exact alternative so it cannot expand into `path`.

Proof is behavioral (public entry points), not regex-source inspection.

### Tests

Table-driven unit coverage for casing/quoting variants, combined multi-key leakage, password regression, and path non-match — all driven through `make_execution_error`.

## Verification

- `todo verify` rungs 1–2: pass
- `uv run -- python -m pytest tests/unit/mcp/test_errors.py -q`: 59 passed
- Behavioral gates for dsn/connection_string/private_key/sas/pat + password regression: pass
- `make ci-lint`: pass
- Full local `make pr-preflight` fast suite: blocked by concurrent hung xdist run at ~95% (chdb native wait; reproduced independently). CI is the authoritative full-suite gate.

## Scope

Only modifies the two allowed paths. Does not touch analytics/benchmark tools, results, or todo_db.

### DIFF
--- a/benchbox/mcp/errors.py
+++ b/benchbox/mcp/errors.py
@@ -44,15 +44,20 @@ _SECRET_SAS_ASSIGNMENT_RE = re.compile(
     flags=re.IGNORECASE,
 )
 # Unquoted PEM / multi-line private keys span whitespace; the generic arm only
-# takes the first token and leaves base64 body lines intact.
+# takes the first token and leaves base64 body lines intact. Truncated PEM
+# (BEGIN without END) and CRLF-separated base64 continuations must still mask.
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1597, opened 2026-08-05
fix(results): filter all platform metadata export boundaries

joeharris76:
## Summary

Close remaining platform-metadata credential egress after #1468.

`build_platform_metadata_payload` already structurally filtered `raw_config`, but `raw_metadata` and the normalized `deployment` / `cloud` / `compute` / `storage` blocks were copied unsanitized into public payloads, private exports, and (when present) results.db. Credential-shaped keys in those mappings could therefore survive every export boundary.

## Changes

- **`environment.py`**: shared `_sanitize_metadata_block` applies `sanitize_platform_options` to every raw and normalized platform mapping at the payload boundary.
- **`database.py`**: when a result carries explicit platform metadata blocks, store a filtered snapshot under `metadata["platform"]` via the same builder (no silent local-DB egress channel).
- **`schema.py`**: comment only — producer path already routes through the builder.
- **Tests**: distinct multi-source sentinels (RAW_CONFIG / RAW_METADATA / DEPLOYMENT / CLOUD / COMPUTE / STORAGE) across public payload, private export (`anonymize=False`), and ResultDatabase; non-secret tuning fields preserved.

## Verification

- Unfixed-tree gate (seq 1): **pass** after fix
- Non-secret preservation (seq 2): **pass**
- `tests/unit/core/results/` fast: **937 passed**, 4 skipped
- Targeted multi-source files: **49 passed**
- `make ci-lint`: **pass** (full `pr-preflight` fast lane hung at ~95% on Ray workers in this environment; scoped results suite used for local proof)

## Scope

Only the paths listed for `credential-egress-platform-metadata-boundary-v3`. Does not touch `platform_options.py` / `anonymization.py` / MCP.

### DIFF
--- a/benchbox/core/results/database.py
+++ b/benchbox/core/results/database.py
@@ -24,6 +24,7 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
 
+from benchbox.core.results.environment import build_platform_metadata_payload
 from benchbox.core.results.models import BenchmarkResults
 from benchbox.core.results.platform_options import sanitize_platform_options
 from benchbox.core.results.regression_policy import is_regression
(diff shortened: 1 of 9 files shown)


joeharris76:
@codex review

PR #1599, opened 2026-08-05
fix(ci): bound develop-post-merge push-drop with sweep + gap detector

joeharris76:
GitHub has dropped push delivery for consecutive develop merges, so
develop-post-merge never started for those SHAs and tip sat un-gated.
Keep the push trigger, add an hourly scheduled sweep of the same gates
against develop tip (fixed concurrency group so sweeps never cancel
push runs), and a read-only daily gap detector that fails when recent
develop SHAs lack any post-merge run.


joeharris76:
@codex review

PR #1601, opened 2026-08-05
fix(results): scrub URI query/fragment credentials with username precision

joeharris76:
## Summary

Scrub credential-bearing URI **query and fragment** parameters in platform options, and keep **username-only** userinfo identity (do not replace non-secret usernames with `****`).

Structural key filtering already redacts options named `access_token` / `password` / `dsn`, but credentials still leaked when those names appeared as URI components on ordinary option keys (export URLs, connection URIs with `sslpassword`, OAuth-style fragments). Username-only authorities were also over-redacted.

## Changes

- `benchbox/core/results/platform_options.py`
  - Userinfo: redact only when a password component is present (`:` in userinfo); preserve username-only and Azure container authorities
  - Query/fragment: mask values of credential-named params via existing `is_secret_option_key` alignment (`access_token`, `sslpassword`, `password`, …)
  - Preserve ordinary hosts, paths, ports, and non-secret params (`x=1`, `application_name`, …)
- Tests covering username-only precision, query/fragment sentinels, and combined userinfo+query cases

## Verification

- TODO rungs 1–2 pass (query/userinfo/ssl sentinels absent; `public-user` and `x=1` preserved; ordinary URLs unchanged)
- `pytest tests/unit/core/results/test_platform_options_capture.py` — 68 passed
- `ruff` / `ty` clean on touched files
- Full `pr-preflight` fast suite hung at ~95% under multi-agent load after `ci-lint` passed; CI will re-run the suite

## Scope

Only:
- `benchbox/core/results/platform_options.py`
- `tests/unit/core/results/test_platform_options_capture.py`

Does not touch `environment.py`, `anonymization.py`, `schema.py`, or `todo_db.py`.

### DIFF
--- a/benchbox/core/results/platform_options.py
+++ b/benchbox/core/results/platform_options.py
@@ -30,6 +30,15 @@ _URI_USERINFO_RE = re.compile(
     flags=re.IGNORECASE,
 )
 
+# Query and fragment parameters. Separators cover the first query ("?"), later
+# query pairs ("&"), the fragment start ("#"), and fragment pairs ("&" after
+# "#"). Credential-named params (access_token, sslpassword, password, ...) are
+# masked; ordinary params (x=1, region=...) are left intact. Name matching
+# reuses is_secret_option_key so key-level and URI-component rules stay aligned.
+_URI_QUERY_OR_FRAGMENT_PARAM_RE = re.compile(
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1602, opened 2026-08-05
fix(todo-db): surface turso auto-provision mint failures

joeharris76:
## Summary

When `turso` is on PATH and auto-provision minting fails, surface a short reason class instead of falling through to the local fallback with no signal.

- `_try_turso_auto_provision` returns `(credentials|None, reason|None)` with classes: `db-show-failed`, `db-show-empty`, `token-create-failed`, `token-empty`, `turso-error`
- Turso absent remains silent `(None, None)`
- `doctor` records `turso-provision` WARN; backend stderr and write-refusal name the class
- Tokens/URLs never printed

Closes `todo-db-auto-provision-mint-failure-visibility`.

## Test plan
- [x] `uv run -- python -m pytest tests/unit/scripts/test_todo_db_turso_autoprovision.py tests/unit/scripts/test_todo_db_doctor.py -q` (31 passed)
- [ ] CI green

### DIFF
--- a/_project/scripts/todo_db.py
+++ b/_project/scripts/todo_db.py
@@ -532,7 +532,7 @@ def _turso_db_name() -> str:
     return os.environ.get("TODO_DB_TURSO_DB") or "benchbox-todo"
 
 
-def _try_turso_auto_provision() -> tuple[str, str] | None:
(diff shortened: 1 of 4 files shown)


joeharris76:
@codex review

PR #1604, opened 2026-08-05
fix(corpus): gate published-results mirror on publication fixed point

joeharris76:
Refuse to build or open a results-data mirror when develop's corpus is not
already at the publication fixed point (re-anonymize would rewrite bytes).

The sync workflow now runs the same check as
test_rederived_corpus_publishes_byte_identically_to_what_is_stored against
develop@GITHUB_SHA before Build mirror branch. Failure is loud and does not
look like a successful no-op. When re-derivation is in flight the mirror waits
on that precondition; once it holds, mirroring proceeds unattended again.


joeharris76:
@codex review

PR #1608, opened 2026-08-05
fix(corpus): union-overlay mirror preserves published-only submissions

joeharris76:
The develop→published-results sync wiped results-data/bundles before
checking out develop, which deleted community submissions that exist only
on published-results whenever develop was also ahead.

Apply a union overlay for bundles instead: checkout develop's files without
git rm -rf of the tree so published-only paths survive. Gate the corpus
drift canary so mixed develop-ahead + published-only never recommends a
wipe-based full mirror.


joeharris76:
@codex review

PR #1610, opened 2026-08-06
fix(results): anonymize nested tuning companion constraint shapes

joeharris76:
List-of-dicts FK tables, slash-delimited local_table values, and
TPC-DI-style references_table/references_column companions leaked raw
identifiers after the simple mapping fix. Recurse into nested dict/list
items under identifier collections and hash additional companion scalar
keys while preserving enabled/action semantics and non-identifier values.


joeharris76:
@codex review

PR #1600, opened 2026-08-05
docs(ops): honest pypi env gate pin (bypass, no canary)

joeharris76:
Soften the pypi required-reviewers section title from (enforced) to
(configured; observed 2026-08-05): GitHub enforces at deploy time, but
unlike the tag ruleset there is no automated environment-drift canary.

Pin can_admins_bypass: true with the accepted single-admin /
prevent_self_review: false posture, and reword drift detection as manual
re-verify against the dated pin.

PR #1611, opened 2026-08-06
chore(ci): bump fast-lane ceiling 27550 -> 28050 for the one-engine tail

joeharris76:
At develop tip 1fb677633e the fast lane collects 27278 against the 27550
ceiling -- headroom 272, just above the 250 floor and shrinking.

The 2026-08-04 bump set that ceiling with 716 headroom explicitly reserved
for the one-engine program's remaining items. ~30 unrelated PRs (#1567..
#1600) consumed 444 of it while none of those items landed. The keystone's
last unit (w4, CLI orchestrator migration onto core/run_service.py) still
adds characterization tests and would cross the ceiling mid-batch.

Standalone ceiling PR per the quantum convention: +500, no test-count
change, one appended log entry. New headroom 772.


joeharris76:
@codex review

PR #1612, opened 2026-08-06
fix(submit): require BENCHBOX_MACHINE_ID_SALT for community publish

joeharris76:
Hard-refuse benchbox submit when the public pseudonym salt is unset so
operators cannot ship empty-salt confirmation-oracle community bundles.
Private/local export without salt remains allowed.


joeharris76:
@codex review

PR #1614, opened 2026-08-06
fix(results): strip residual empty client_host from the public corpus

joeharris76:
Omit hollow client_host maps at the public boundary and re-derive 105
primary bundles so stored bytes match the fresh public shape.


joeharris76:
@codex review

PR #1615, opened 2026-08-06
fix/staging manifest certifies stale data

joeharris76:
- **fix(core): rebuild stale staging data instead of certifying it**
- **fix(core): invalidate manifests from the certify-without-rebuild generation**


joeharris76:
@codex review

PR #1616, opened 2026-08-06
fix(uat): surface unvalidated cells in every machine-readable roll-up

joeharris76:
PR #1577 added SubmitTerminalState.unvalidated and stopped the package
phase treating a never-validated cell as a failure, which was the
headline bug: a full DataFrame sweep read as 0% pass. It stopped at the
classifier and the package phase.

Three roll-up surfaces are built from the same ReportSummary and none of
them knew the state existed, so an unvalidated cell was folded silently
into `passed` with no adjacent disclosure:

  - matrix_summary.tsv's footer (human-readable)
  - uat_gate_summary.json's accounting block (what a release gate reads)
  - `uat report --json`'s summary object (what tooling reads)

A DataFrame release-gate stage could therefore be majority-unvalidated
and every artifact would report an ordinary clean pass. That is a worse
failure than the one #1577 fixed: the first was loud and wrong, this one
is quiet and wrong.

ReportSummary gains unvalidated_count and PhaseAccounting gains
unvalidated, both threaded through orchestrator._accounting_for_gate_summary
in both its branches. The count is cross-cutting, NOT a disjoint bucket:
an unvalidated cell is already counted in passed/attempted, and the count
deliberately does not feed exit_code() -- UAT must not treat unvalidated
as a cell failure.

Mutation-checked: with the four source files reverted, six of the added
tests fail.

TODO: unvalidated-results-misclassified-as-schema-violations-20260805


joeharris76:
@codex review

PR #1617, opened 2026-08-06
fix(mcp): preserve benign secret prose in error scrubbing

joeharris76:
Bare-prose matches after secret keys ("password X", "token Y") were
scrubbing ordinary diagnostic English such as "secret sauce", "token
refresh", and "password reset". Keep connector and assignment forms
aggressive; only scrub bare-prose tokens that look credential-like.

Depends on #1595 (vocabulary / SAS / PEM scrub arms).


joeharris76:
@codex review

PR #1618, opened 2026-08-06
fix/uat compose project name budget

joeharris76:
- **fix(uat): derive compose project-name budget from the container-id limit**
- **fix(uat): correct compose rename comment, close lakesail exemption gap**


joeharris76:
@codex review

PR #1626, opened 2026-08-06
fix/lakesail velox compose env fallback

joeharris76:
- **fix(docker): require BENCHBOX_DATA_DIR instead of nesting compose defaults**
- **fix(docker): use per-stack .env fallback instead of mocker-unsafe compose forms**
- **fix(docker): require an absolute BENCHBOX_DATA_DIR, no compose default**
- **fix(docker): validate BENCHBOX_DATA_DIR at every bring-up/teardown path**


joeharris76:
Review findings landed in `56a383bf5b`.

**F2 (HIGH — regression this PR introduced) is fixed.** The guard sat below `trap cleanup EXIT INT TERM`, so its own `exit 1` fired the failure-cleanup trap, running `compose down -v` and deleting the tracking file. A second shell running the same idempotent bring-up without the export would tear down a *live* stack from the first shell and orphan it. Guard is now hoisted above both the trap and `status=1`.

Also landed: F1 (Makefile guard coverage), F3 (parity-script DATA_DIR wiring coverage), F4 (`compose_environment` no longer returns `{}` for a `None` `benchmark_runs_dir` on lakesail/velox, which let the child inherit ambient `BENCHBOX_DATA_DIR` via `os.environ.copy()`).

### Mutation verification

Each mutation is deliberately **weaker than a full revert**, and each was confirmed to fail for the *right* reason — non-zero-exit and tracking-file assertions, not incidental `ImportError`/`ValueError`:

| Mutation | Before | After |
|---|---|---|
| `M2a` — guard deleted from `test-docker-up-%` only, kept in `test-docker-%` | **SURVIVED** 748/748 | **KILLED** — 5 rungs fail on `returncode == 0` |
| `F2` — guard restored to pre-fix placement (present, just below the trap) | n/a | **KILLED** — `test_guard_rejection_leaves_a_tracked_stack_untouched` fails with *"guard rejection must not delete the tracking file for an already-running stack"*, i.e. the exact operator symptom |
| `M4` — the three `BENCHBOX_DATA_DIR` pass-throughs stripped from the parity script | **SURVIVED** | **KILLED** — 3 rungs fail |

The `F2` mutation is the one that matters most here: `M2a`-style deletion is a strong mutation that a weak test could still catch, whereas F2 is a pure *placement* bug. The rung distinguishes them.

Tests shell out to the real `make`/real script with a stub `CONTAINER_ENGINE`/`make` on `PATH` that only logs argv and exits 0 — the guard is genuinely exercised and its *absence* of any compose call is proven, without touching a real container engine.

Full suite: **27,947 passed, 24 skipped**.

Not touched, deliberately: `test_no_repo_env_file_under_docker_sets_a_non_absolute_benchbox_data_dir` stays green on the unfixed tree because no `.env` exists under `docker/`. It is a valid forward guard, just not a rung for this fix.

### Note on `_project/specs/uat-framework.md`

This PR still carries a regenerated LOC-table diff in that file. #1634 removes that generated block entirely — it is the sole merge conflict across #1618/#1627/#1631 and the only one here. Once #1634 lands, this PR needs one final resolution taking develop's side of that file (this PR contributes only regenerated numbers to it — verified), after which the conflict class is gone.

joeharris76:
@codex review

PR #1627, opened 2026-08-06
fix/uat disk budget platform chunking

joeharris76:
- **feat(uat): model per-platform disk budget and add execute.platform_chunking**
- **fix(uat): wire the platform-chunking disk gate into live preflight**
- **fix(uat): withdraw platform_chunking; disclose the disk budget as a lower bound**


joeharris76:
@codex review

PR #1630, opened 2026-08-06
fix/uat cell teardown orphans

joeharris76:
- **fix(uat): reap orphaned child process groups on sweep cancellation**
- **fix(uat): review follow-up on child-process-group teardown fix**
- **fix(uat): complete the kill ladder under a second signal; retarget its tests**


joeharris76:
@codex review

PR #1631, opened 2026-08-06
fix/uat readiness and memory gate

joeharris76:
- **fix(uat): make the free-memory gate real, wired, and covered**
- **fix(uat): detect a stack that dies mid-platform, not just one that never started**


joeharris76:
## Review: readiness + memory gate — verdict **PASS**, no blocking findings

Reviewed against the charter question: *does the per-cell probe actually catch the incident — platform up, readiness passes, stack dies partway through the cell list — and do the remaining cells land in `died_mid_platform` rather than as ordinary cell failures?*

**Yes, and it is properly pinned.** Seven mutations, all KILLED, three of them on the consumer side.

### The mechanism holds up

`_run_platform_benchmark` probes immediately before handing each cell to the runner, so the covered failure is death at *arbitrary* latency, not just at startup. On a probe failure it raises `_PlatformDiedMidRun` with `pb_cells[index:]` — including the cell that was about to run — and `_run_or_skip_platform` adds every later benchmark of the same platform. Using an exception rather than a return flag is right: it skips the interrupted pair's `completed_pairs.add(...)` tail, so database pruning cannot treat a half-run pair as finished.

On the timing concern from the handoff (the real CedarDB incident died ~29 s in, past develop's one-shot ~10 s window): the tests simulate death by *probe ordinal* (`_probe_dying_after(3)` — readiness, arm, first cell, then dead) rather than wall clock. That is the correct axis, not a weaker one: the probe runs before every cell regardless of elapsed time, so cell-ordinal is the variable the logic actually depends on. It is not an immediate-crash simulation — the first cell runs and passes before the stack dies.

### Mutation results

| Mutation | Result |
|---|---|
| `pb_cells[index:]` → `pb_cells[index+1:]` (the cell it died on silently dropped) | **KILLED** — 2 rungs, incl. the `1 run + 3 died == 4 defined` conservation assert |
| Later benchmarks of the same platform not claimed | **KILLED** |
| `exit_code()` no longer treats a lost platform as a failure | **KILLED** — this is the "quiet-and-wrong shape of the original bug, just relocated" |
| A consumer's `getattr(execute_outcome, "died_mid_platform", ())` name goes stale | **KILLED** by `test_mid_platform_death_surfaces_in_every_machine_readable_rollup` |
| Report stops flagging a lost platform as needing attention | **KILLED** — 2 rungs |
| `-a` dropped from `compose ps` | **KILLED** — 2 rungs, incl. consumer-side `test_execute_readiness_check_requests_ps_all` |
| Empty `ps` table treated as healthy | **KILLED** — `test_execute_readiness_check_fails_closed_on_an_empty_compose_ps_table` |

I specifically probed the batch's dominant defect class — a rung pinning the SOURCE of a behaviour but not its CONSUMERS. The pervasive `getattr(..., default)` calls were the obvious place for it to hide, since a stale name degrades silently to `0`. It is covered.

### Confirmed: the `-a` fix closes a live bug in `develop`

Independently verified on `develop@3ea0defdf5`, and it is worse than "missing a flag" — **both halves fail open**:

1. `compose_ps_command` builds a bare `ps`. Docker Compose lists only *running* containers, so a service that started and then exited is simply **absent** from the table.
2. `compose_ps_unhealthy_services` is documented fail-open: *"Empty stdout, a header-only table, or a table with only 'Up ...' rows all return `()` -- readiness passes."*

Composed, develop's post-start readiness check **cannot detect the exact "started, then exited" case its own docstring says it exists to catch**. It happens to work under mocker (whose `ps` does list exited containers), which is presumably why it survived — but the repo supports both engines and the gate silently no-ops on real Docker.

This branch fixes both halves and adds a wider not-ready vocabulary. Worth landing regardless of what happens to the rest of the PR.

### Verification

Full suite on the branch: **27,924 passed, 24 skipped**. Zero conflict markers worktree-wide. Working tree clean apart from one untracked handoff note.

### Not blocking

- The only conflicting file is `_project/specs/uat-framework.md`; `orchestrator.py` and `test_sweep_durability.py` auto-merge. #1634 removes that conflict class. This PR contributes only regenerated numbers to that file (verified), so resolve by taking develop's side once #1634 lands.
- Per the batch handoff, auto-merge/label state is left alone pending the auto-merge policy ADR. The `no-auto-merge` label is untouched.

joeharris76:
@codex review

PR #1632, opened 2026-08-06
docs(ops): add develop push-drop coverage inventory

joeharris76:
Companion to develop-post-merge-gaps.md, which covers the failure class
(GitHub can silently drop push delivery for consecutive develop merges)
and the backstop built for develop-post-merge.yml specifically.

This inventories every *other* workflow whose on.push can fire for a push
to develop: whether it already has a schedule backstop, its role
(safety-critical vs advisory), the residual push-drop risk, and the
disposition. Two subjects lack a schedule by design:

- sync-results-data-to-published.yml -- accepted risk; the daily read-only
  corpus-drift-check.yml canary bounds detection rather than putting a
  write-capable public-branch mirror on cron.
- results-explorer-browser.yml -- accepted risk; the pre-merge required
  check is the safety property, and an hourly Playwright suite is the
  expensive pattern this inventory rejects.

Also records the enumeration method, explicit non-entries, cost/policy
constraints, and a manual recovery cheatsheet.

Tests: a link assertion on the gaps doc, a subject-naming assertion on the
inventory, and a drift guard that re-derives the develop-push subject set
from .github/workflows/ and fails if it diverges from the table.

### DIFF
--- a/docs/operations/develop-post-merge-gaps.md
+++ b/docs/operations/develop-post-merge-gaps.md
@@ -34,6 +34,11 @@ develop-post-merge.yml` `headSha`.
 That is a silent failure mode. Push-only guards share it. Every workflow
 that only fires on `push` to develop can miss the same window.
 
+**Full inventory of develop-push workflows** (which ones already have a
+schedule, which residual gaps are accepted, and recovery commands) lives in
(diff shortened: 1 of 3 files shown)


joeharris76:
@codex review

PR #1633, opened 2026-08-06
fix(ci): require every develop-ruleset context in green-unmerged sweep

joeharris76:
## Problem

The nightly green-unmerged sweep decided "required lane green" from the latest `ci-required-result` check run alone. Since 2026-08-03 the `develop-squash-only` ruleset has **two** required status contexts ([`docs/operations/repo-admin-settings.md`](docs/operations/repo-admin-settings.md)):

```text
- ci-required-result
- Results Explorer browser gate
```

So a PR green on the umbrella while `Results Explorer browser gate` was red — or had never reported — was classified green. Combined with the other strand predicates it could be alerted as stranded when it was in fact correctly unmergeable. Fetching is unfiltered (`commits/{sha}/check-runs`), so the gate's run was always present in the data; only the classifier ignored it.

## Change

- `latest_check_run(runs, name)` generalizes the old single-name lookup, keeping the newest-`started_at` wins rule that stops a stale re-run conclusion from winning.
- `is_required_lane_green` now requires **every** context in the new `REQUIRED_CHECK_NAMES` tuple to be latest-success. Missing runs stay fail-closed not-green — the browser gate always reports on develop PRs (its path-aware skip still concludes `success`), so a silent context is a real signal rather than an expected absence.
- `required_lane_check_run()` is **dropped**, not kept as a compat shim: nothing called it once `is_required_lane_green` stopped doing so, and its `REQUIRED_CHECK_NAMES[0]` indexing made tuple order load-bearing when the order carries no meaning.
- [`docs/operations/pr-triage.md`](docs/operations/pr-triage.md) now states what "required lane green" means, which was ambiguous next to `soundness-drain.md`'s single-context wording.

## Verification

```bash
uv run -- python _project/scripts/green_unmerged_sweep.py --self-test
uv run -- python -m pytest tests/unit/scripts/test_green_unmerged_sweep.py -q -n 0
```

- Self-test: 12 fixture PRs, 3 stranded (`2008/2009/2010` — unchanged from before).
- Unit tests: 61 passed. Full `tests/unit/scripts/` 1615 passed; `test_ruleset_drift_review_coverage.py` 26 passed.
- The ruleset pin parses the required contexts from the same runbook block [`scripts/ruleset_drift_check.py`](scripts/ruleset_drift_check.py) reads, rather than restating them as a second hardcoded copy that could drift silently.
- New fixture PRs `2011`/`2012` cover partial green (gate absent, gate red) as end-to-end `not_green` exclusions. Existing fixture PRs each gained a passing gate run so the case each was written to pin stays the only variable — verified additive-only against the previous fixture.
- Negative-tested: reverting the constant to a single context fails the drift pin, both partial-green regressions, and the self-test.

## Scope / guardrails

Follows the scope rules of the (already-complete) tracker item `auto-merge-verify-and-triage-runbook-2` that introduced this sweep in #1288 — all touched paths fall under its `only_modify` globs (`_project/scripts/**`, `tests/unit/scripts/**`, `docs/operations/**`); no `do_not_modify` path touched. Its preserved invariants hold: soundness-gated PRs are still excluded, `auto-merge-on-open.yml` is untouched, and the sweep remains alert-only — it still never enables auto-merge.

### DIFF
--- a/_project/scripts/fixtures/green_unmerged_fixture.json
+++ b/_project/scripts/fixtures/green_unmerged_fixture.json
@@ -12,9 +12,22 @@
       "head_pushed_at": "2026-07-23T08:00:00Z",
       "auto_merge": null,
       "timeline_events": [],
-      "changed_files": ["benchbox/platforms/duckdb/adapter.py"],
+      "changed_files": [
+        "benchbox/platforms/duckdb/adapter.py"
+      ],
       "check_runs": [
-        {"name": "ci-required-result", "status": "completed", "conclusion": "success", "started_at": "2026-07-23T08:10:00Z"}
+        {
+          "name": "ci-required-result",
+          "status": "completed",
+          "conclusion": "success",
+          "started_at": "2026-07-23T08:10:00Z"
+        },
+        {
+          "name": "Results Explorer browser gate",
+          "status": "completed",
+          "conclusion": "success",
+          "started_at": "2026-07-23T08:12:00Z"
+        }
       ]
     },
     {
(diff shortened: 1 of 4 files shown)


joeharris76:
@codex review

PR #1634, opened 2026-08-06
fix(uat-spec): replace the committed LOC table with a ceiling gate

joeharris76:
## Problem

`_project/specs/uat-framework.md` is currently the **sole** merge conflict for #1618 and #1627, and the only conflicting file for #1631 (their other files auto-merge). Verified with `git merge-tree` against `develop@3ea0defdf5`.

The cause is #1258 (`593a150b78`), which fixed a *staleness* problem — the manual LOC table drifted, hand-corrected twice in a week — by committing **exact `wc -l` counts** plus a generated per-bucket summary, checked byte-for-byte by pre-commit and CI.

That trade has a measurable cost. Every commit touching the spec since #1258 is **100% regenerated numbers, zero prose** — eleven in a row:

```
411f874bcb #1630 gen=10/10    2bf654c977 #1628 gen=18/18
bf35098399 #1616 gen=14/14    5ccce7180a #1591 gen=6/6
b946f47cc9 #1577 gen=6/6      c0c96bddbd #1316 gen=6/6
```

And because the per-bucket summary is a handful of **shared** lines, PRs touching completely disjoint UAT modules still collide there — then collide again on *every replayed commit* during a rebase.

## Reframe

The table exists to answer a charter question: *has UAT grown out of scope?* That is answered by **magnitude**, not by an exact line count. The committed precision (±1 line) vastly exceeds the decision it informs, and buys a conflict per PR.

## Change

Stop committing the numbers; commit **ceilings**.

- `_project/specs/uat-loc-budget.json` — per-bucket + total ceilings.
- `uat_loc_table.py --check` measures the tree and fails only on a breach. Bare invocation prints current numbers and headroom, and writes nothing.
- The spec drops the volatile `Actual` column and the generated block, and documents the budget instead.
- Kept: the bucket-map ↔ spec-table coverage guard. Added: a guard for bucket labels drifting between the JSON and `BUCKETS`.

An ordinary UAT change now moves a number nobody stores → no spec diff, no conflict. The file changes only when someone deliberately raises a ceiling — which is exactly the charter decision the table existed to prompt, now explicit and enforced rather than eyeballed in a doc that was stale by default.

## Prior art

| Path | Concept | Decision |
|---|---|---|
| `_project/evals/agent-instructions/scenarios.json` | committed `budgets` vs measured actuals | **extend** — same pattern |
| Ruff C901 cap (18) | committed threshold, computed actual | **reuse** |
| `_project/scripts/uat_loc_table.py` (#1258) | exact-value drift check | **supersede** |

This also implements the filed-but-unstarted tracker item `set-a-ci-loc-budget-guard-that-fails-when-uat-exceeds-the-ag`, whose note already specified this design and marked it buildable once #1258 merged.

## Ceilings

Set to clear the entire in-flight batch. Max observed per bucket across #1618/#1626/#1627/#1631 sums to **11,611** vs develop's 11,043; ceilings sit ~15% above that (total **13,500**, current utilisation 81.8%). Landing the batch cannot trip the gate.

## Not a regen guard any more

Removed from `make guards-fix` and from the regen table in `docs/operations/guards-fix.md`. The standing assumption in this repo is "check guard failed → run guards-fix"; here that is wrong, so the failure message says so explicitly and `test_guard_messages.py` pins that denial.

## Verification

Three mutations, all **KILLED**:

| Mutation | Result |
|---|---|
| Real LOC breach (`throughput.py` +200 lines) | exit 1, `OVER BUDGET: throughput is 516 lines, ceiling 400 (+116)` |
| Bucket label renamed in the JSON | rejected, names both sides of the drift |
| Module dropped from the spec table | rejected, names `ladder.py` |

Plus a rung asserting the committed ceilings hold for the tree as shipped. `ruff check`/`format` clean; full pre-commit suite passed on commit.

## Consequence to accept

At-a-glance per-module numbers leave the doc. A number in a committed doc is stale-by-default and costs a conflict per PR; the script prints current numbers on demand.

## Sequencing

Worth landing **before** the remaining UAT PRs. Each will need one final conflict resolution against this change (take develop's deletion), then the conflict class is gone rather than paid per-PR and per-replayed-commit.


joeharris76:
@codex review

PR #1635, opened 2026-08-06
fix(ci): require every develop-ruleset context in soundness drain report

joeharris76:
Sibling of #1633, same class of defect in the other script that classifies the develop required lane.

## Problem

`_project/scripts/soundness_drain_report.py` carried its own copy of the single-context logic: `REQUIRED_CHECK_NAME = "ci-required-result"`, matched by `required_lane_check_run()`, with `is_required_lane_green()` returning True from that one run. The `develop-squash-only` ruleset has had **two** required status contexts since 2026-08-03 ([`repo-admin-settings.md`](docs/operations/repo-admin-settings.md)):

```text
- ci-required-result
- Results Explorer browser gate
```

So a PR green on the umbrella with the browser gate red — or never reported — was classified required-lane green, and the report could label it `awaiting-owner` and file it in the drain digest as "parked on the owner" when it was in fact correctly unmergeable.

## The second bug, not in the original report

`park_time_hours()` anchored *"when did this PR go green"* on the umbrella's `completed_at`. The lane is only green once **every** required context has finished, so anchoring on the earlier one overstates park time whenever the browser gate completes later — tripping the >24h drain gate early and generating a false "parked 2 days" alert.

Fixture PR `9110` pins exactly this: umbrella finished 52h before `as_of`, gate 10h before.

| anchor | park time | verdict |
|---|---|---|
| umbrella only (old) | 52.0h | qualifies — **false alert** |
| last context (new) | 10.0h | too young — correct |

Fixing the classifier without the anchor would have left this live.

## Change

- `latest_check_run(runs, name)` + `is_check_run_success()` replace the single-name lookup; `is_required_lane_green()` requires **every** context in the new `REQUIRED_CHECK_NAMES` tuple to be latest-success, missing runs fail closed.
- `required_lane_green_at()` returns the `completed_at` of the last required context to finish, falling back to `updated_at` only when a context lacks a stamp.
- [`docs/operations/soundness-drain.md`](docs/operations/soundness-drain.md) updated for both the green definition and the park-time anchor.

## Tests

This script had **no unit test file** — only the fixture-driven `--self-test`. Adds `tests/unit/scripts/test_soundness_drain_report.py` (32 tests): required-context predicates, park-time anchoring, partial-green classification. The ruleset pin parses the contexts from the same runbook block [`scripts/ruleset_drift_check.py`](scripts/ruleset_drift_check.py) reads rather than restating them.

```bash
uv run -- python _project/scripts/soundness_drain_report.py --self-test
uv run -- python -m pytest tests/unit/scripts/ -q -n 0
```

- Self-test: 10 fixture PRs, 3 qualifying — `9101/9105/9107`, unchanged.
- `tests/unit/scripts/`: 1628 passed.
- New fixture PRs `9108`/`9109` (gate absent, gate red) as `not_green`; `9110` for the anchor.
- Existing fixture PRs each gained a gate run completing *before* their umbrella, so the park anchor and every prior expectation are untouched — verified additive-only against the previous fixture.
- Negative-tested: reverting only the anchor makes `9110` wrongly qualify, caught by both the self-test and the unit test.

## Why no shared helper yet

The obvious dedup — one helper module for both scripts, alongside the existing `auto_merge_soundness_paths.py` precedent — is deliberately **not** in this PR. `green_unmerged_sweep.py` is in flight in unmerged #1633, so editing it here would either conflict on merge or require a stacked PR, which gets zero CI in this repo. Filed as a follow-up to do once #1633 lands. Until then both sides independently pin to the runbook, so they agree transitively rather than through a merge-order coupling.

Tracker: `soundness-drain-all-ruleset-required-checks-20260806`.

### DIFF
--- a/_project/scripts/fixtures/soundness_drain_fixture.json
+++ b/_project/scripts/fixtures/soundness_drain_fixture.json
@@ -10,7 +10,10 @@
     "9102": "auto_merge_on",
     "9103": "too_young",
     "9104": "not_green",
-    "9106": "draft"
+    "9106": "draft",
+    "9108": "not_green",
+    "9109": "not_green",
+    "9110": "too_young"
   },
   "prs": [
     {
@@ -33,6 +36,12 @@
           "status": "completed",
           "conclusion": "success",
           "completed_at": "2026-07-16T08:00:00Z"
+        },
+        {
+          "name": "Results Explorer browser gate",
+          "status": "completed",
+          "conclusion": "success",
+          "completed_at": "2026-07-16T07:55:00Z"
         }
       ]
     },
@@ -59,6 +68,12 @@
           "status": "completed",
           "conclusion": "success",
           "completed_at": "2026-07-15T08:30:00Z"
+        },
+        {
+          "name": "Results Explorer browser gate",
+          "status": "completed",
+          "conclusion": "success",
+          "completed_at": "2026-07-15T08:25:00Z"
         }
       ]
     },
@@ -82,6 +97,12 @@
           "status": "completed",
           "conclusion": "success",
           "completed_at": "2026-07-23T09:45:00Z"
+        },
+        {
+          "name": "Results Explorer browser gate",
+          "status": "completed",
+          "conclusion": "success",
+          "completed_at": "2026-07-23T09:40:00Z"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


joeharris76:
@codex review

PR #1637, opened 2026-08-06
fix(uat): pin run_cell's timeout consumer, and close the orphans follow-ups

joeharris76:
#1630 merged before its review completed. Its verdict on the blocking R1 question came back ACCEPTABLE, so the merge was fine — but the review found a defect that is **live in develop right now**, plus five lower-severity items. This closes all six.

## F1 — the fix's sole consumer was unpinned

#1630's process-group teardown has exactly one production consumer: `runner.run_cell`. Nothing pinned that wiring.

- Every test in `tests/uat/test_runner.py` replaces `runner.run_with_timeout` with a fake — **18 call sites**.
- The "end-to-end wiring" test in `tests/uat/test_sweep_durability.py` calls `timeouts.run_with_timeout` *directly*, bypassing `run_cell`.

So re-pointing runner's import at a pre-#1630 implementation left the original defect **fully live**, with **600 tests passing and 0 failing**. This is the batch's dominant defect class — a rung that pins the SOURCE of a behaviour but not its CONSUMER.

`tests/uat/test_runner_timeout_process_group.py` closes it from the other side: `run_with_timeout` stays **real** and the *argv* is substituted, so the assertion exercises the exact call `run_cell` makes. Two rungs — a timed-out cell must leave no surviving grandchild, and the direct child must be its own session leader (the precondition that makes `killpg` safe rather than sweep-suicidal).

## F2–F6

| | Finding | Fix |
|---|---|---|
| **F2** | `except BaseException` asserted in prose, but every cancellation rung raised a `KeyboardInterrupt` subclass, so narrowing the guard survived | rung raising a bare `BaseException` (the `GeneratorExit`/`SystemExit` shape) |
| **F3** | R1 docstring claims the residual leaves "one visible process" — measured, 18/18 survivors already owned a live grandchild. **The R1 acceptance rests on this docstring.** | reworded to one recoverable session-leader *root*, reapable with a single `kill -- -PGID` |
| **F4** | gate 2 is only meaningful while the 20 ms probe tick ≪ the 200 ms grace window; assumed, never asserted | static margin assertion — no timing, no flake |
| **F5** | `_REAP_TIMEOUT_S` introduced but the timeout path's drain kept a hardcoded `1.0` | applied, and pinned by patching the constant to a value no literal matches |
| **F6** | `test_sweep_durability.py` misattributed the call site to `exec_phase.run_cell`; `execute.py` never imports it | corrected, and the docstring now states what that test does *not* cover |

## Verification

Every mutation weaker than a full revert, each confirmed to fail for the right reason:

| Mutation | Result |
|---|---|
| Consumer re-pointed at a pre-#1630 implementation, `timeouts.py` untouched — **the review's own mutation** | **KILLED** by both new rungs (was 600 passed / 0 failed). Fails in 92 s with a precise diagnostic |
| `except BaseException` → `except KeyboardInterrupt` | **KILLED**, and only by the new F2 rung (1 failed, 13 passed) |
| Drain reverted to hardcoded `1.0` | **KILLED** by the new F5 rung |
| `_KILL_GRACE_S` shrunk to the probe tick | **KILLED** by the new F4 rung |

Full suite: **27,913 passed, 24 skipped**. `ruff check`/`format` clean.

### One negative result worth recording

An earlier draft of the F1 rung **passed against the broken implementation, for the wrong reason**. Signalling only the direct child leaves the grandchild holding the inherited stdout pipe, so `communicate()` blocked for the child's full lifetime — by which point the grandchild had exited on its own, and "is it dead?" found a corpse the code under test had not caused. The 300-second runtime was the only tell.

Fixed by making the spawned lifetime far exceed the test's own ceiling, and by bounding the call on a worker thread — which also converts a *hang* into a *verdict*. Both changes are commented in the module, since the trap is easy to reintroduce.

## Note on `_project/specs/uat-framework.md`

That hunk is pure regenerated LOC noise triggered by a 5-line docstring/constant change — a live instance of the churn #1634 removes. It is also why this PR will conflict there until #1634 lands.


joeharris76:
@codex review

PR #1639, opened 2026-08-06
fix(ci): stop the submission gate failing open on a failing validator

joeharris76:
## The bug

`Validate bundles` pipes `scripts/validate_submission.py`'s stdout into `tee`. A bash pipeline returns the status of its **last** command, and GitHub Actions' default shell for `run:` is `bash -e {0}` — `-e` but **not** `-o pipefail`. This workflow declares no `defaults.run.shell`, so `tee`'s zero exit masked the validator's non-zero one completely.

Found on [#1629](https://github.com/joeharris76/BenchBox/pull/1629), a maintainer corpus mirror into `published-results`. The validator posted:

> ## Submission Validation: FAILED
> Validated **105** bundle(s).

**105 of 105 bundles failing** — every one reporting both `summary.validation must be 'passed' for public submissions, got 'partial'` and a bundle hash mismatch. The required check went green anyway and the PR reported `MERGEABLE / CLEAN`.

## Why this is worth its own PR

The blast radius is not one PR. This is the gate on **every** submission, community ones included, and it has been reporting green regardless of the validator's verdict.

It also silently disarmed the step after it — `Check corpus inventory` runs `if steps.validate.outcome == 'success'`, a condition that had become permanently true.

## The fix

`set -euo pipefail` in the step.

I swept every workflow for a `| tee` inside a `run:` block without pipefail. The only other two sites are in `release-canary.yml`, and both already handle it correctly with `|| rc=${PIPESTATUS[0]}`. This was the sole defective site — and the repo already had a house idiom for it.

The comment above the invocation claimed *"exit code reflects validation result"*, which was false. Corrected, and the reason the pipefail line is load-bearing now sits where someone might otherwise delete it as hygiene.

## Regression rung

`tests/unit/workflows/test_validate_submission_fail_open.py` executes the **real** `run:` block lifted from the workflow under a real shell, substituting only the validator invocation with a stub of known exit status.

Asserting that `"pipefail"` appears in the YAML would have been the weaker test: it pins the current *spelling* of the fix rather than the behaviour, and would survive a later restructuring that loses the status some other way.

The validator call is stubbed rather than executed because `xargs -d` is GNU-only and would make the rung red on macOS for a reason unrelated to this gate — the "red for the wrong reason" trap.

## Verification

| Mutation | Result |
|---|---|
| `set -euo pipefail` removed (reverts to the #1629 state) | **KILLED** — step exits 0 with a failing validator, rung fails with that exact diagnostic |
| Paired passing-validator rung | stays **green** — the fix is not a green-to-red overcorrection that would block every legitimate submission |

`tests/unit/workflows` + `tests/unit/scripts`: **1,834 passed**. YAML validates. Full pre-commit suite passed on commit.

## Out of scope

This does **not** address *why* #1629's 105 bundles fail. The `partial` validation status and the universal manifest hash mismatch are a separate question deserving their own change — plausibly two separate ones. This PR only ensures such a failure is visible instead of green.

### DIFF
--- a/.github/workflows/validate-submission.yml
+++ b/.github/workflows/validate-submission.yml
@@ -250,6 +250,18 @@ jobs:
           IS_FORK: ${{ github.event.pull_request.head.repo.fork }}
           HEAD_REF: ${{ github.head_ref }}
         run: |
+          # `pipefail` is load-bearing, not hygiene. The validator's exit code is
+          # this gate: it is piped into `tee` below, and a bash pipeline returns
+          # the status of its LAST command. GitHub's default shell is
+          # `bash -e {0}` -- `-e` but NOT `-o pipefail` -- so without this line
+          # `tee`'s zero masked a failing validator entirely. Observed on PR
+          # #1629: the validator reported "Submission Validation: FAILED" with
+          # 105/105 bundles failing, this step still exited 0, and the required
+          # check went green. It also silently disarmed the corpus-inventory step
+          # below, which is gated on `steps.validate.outcome == 'success'`.
+          # Do not remove `| tee` without keeping the exit code, and do not
+          # remove this line while the pipe exists.
+          set -euo pipefail
           REQUIRE_MANIFEST="--require-manifest"
           if [ "$IS_FORK" = "false" ]; then
             case "$HEAD_REF" in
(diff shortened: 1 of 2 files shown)


joeharris76:
@codex review

PR #1760, opened 2026-08-17
fix(ci): accept skill-plus-content preflight

joeharris76:
## Summary

Follow-up for review comments on already-merged #1759.

- Accept the classifier's valid `skill_integrity_only=true` + `content_guard_needed=true` decision and run the two narrow lanes (`skill-integrity` + `pr-content-guard`) instead of failing closed.
- Keep `make pr-preflight-fast-tests` self-contained so the opt-in `BENCHBOX_PREPUSH=1` hook still works when no parent classifier artifacts are supplied.

## Test plan

- [x] Focused routing tests for skill-only, skill-plus-content, mixed, product/unknown/empty, malformed, and missing-tool cases
- [x] Direct `pr-preflight-fast-tests` invocation recreates classifier artifacts
- [x] Ancestry / `pr-open` / `pr-refresh` / `pr-fanout` safety pins remain unchanged
- [x] 179 local-preflight / parity / release-infrastructure tests passed


joeharris76:
Fixed in 9838e5fba: regenerated \`make/inventory.json\` via \`python make/check_makefile_inventory.py --write\` to pick up the \`.pr-preflight-route\` and \`pr-preflight-fast-tests\` recipe changes. Drift guard now passes.

PR #1761, opened 2026-08-17
fix(uat): model per-platform disk demand and add chunked execute

joeharris76:
Preflight now sums measured loaded-database rows per platform and names
the concurrent vs chunked envelopes. A shortfall aborts with the computed
gap instead of exhausting disk mid-sweep. execute.platform_chunking walks
one platform at a time and prunes through the reuse-aware helpers.


joeharris76:
Fixed in 557717236: \`check_disk_headroom\` now takes a \`platform_chunking\` flag and gates on \`chunked_database_gib\` (max, one platform at a time) instead of \`concurrent_database_gib\` (sum, all platforms) when \`execute.platform_chunking\` is enabled. Added a regression test pinning both directions.

PR #1762, opened 2026-08-17
docs/review protocol axes 20260817

joeharris76:
- **docs: add architecture and plan review axes**
- **docs: restore REVIEW-AUTH later-turn wording for the audit fixture**


joeharris76:
Fixed in a19a57f79: corrected the synchronize-workflow inventory — removed Documentation (path-filtered by `docs.yml`, does not start on a `tests/**`-only PR) and added the unconditional `develop-refresh-shadow` workflow.

PR #1939, opened 2026-08-29
chore/corpus trust cut 20260823

joeharris76:
Withdraws every result run before 2026-08-23 from the curated corpus, and adds one fresh run so the cohort floor still holds.

## Why

The corpus was 138 bundles, **130 of them predating 2026-08-23**:

| Run date | Bundles | Origin |
|---|---|---|
| 2026-04-03/04 | 12 | Original corpus generation |
| **2026-05-02** | **111** | Maintainer UAT sweep (#164) |
| 2026-05-12 | 3 | JoinOrder canonical UAT |
| 2026-07-30 | 4 | SSB seed-lane runs |
| 2026-08-23/25 | 8 | TPC-DS SF1/SF10 (#1839) + TPC-H SF1 DataFrame pair |

They persisted because nothing in the pipeline has an age, freshness, or retention concept. Bundles had only ever been removed for soundness reasons, never for age.

**This is a trust decision, not a soundness finding.** No defect was found in the removed results. Unlike the 2026-07-16 tuned drop (#1176 proved the tuning config never reached adapters) and the 2026-08-24 zero-query withdrawal (bundles claimed `passed` after executing nothing), there is no bug report behind this one. `CORPUS_NOTES.md` states that explicitly so a future reader does not go hunting for one.

## What changes

**`a39d80cc9` — add TPC-H SF1 DuckDB.** After the cut, TPC-H SF1 would hold only Polars and PySpark, failing the `>=3`-platform floor in `validate_corpus.py`. This adds a SQL reference point against two DataFrame-mode results, and anchors the home page's cross-benchmark aggregation since DuckDB also appears in both TPC-DS cohorts. DuckDB 1.3.2, 66 queries, 0 failed, validation phase PASSED. No manifest sidecar, so the inventory derives `trust_label=maintainer-run`.

**`223acbf48` — the withdrawal.** 130 bundles + 114 sidecars (244 files), plus inventory, README, and `CORPUS_NOTES.md` updates.

Surviving corpus — 9 bundles, 3 cohorts:

| Benchmark | Scale | Platforms |
|---|---|---|
| tpcds | 1.0 | DataFusion, DuckDB, Spark |
| tpcds | 10.0 | DataFusion, DuckDB, Spark |
| tpch | 1.0 | DuckDB, Polars, PySpark |

## One test needed changing

`test_every_manifest_bundle_hash_matches_its_bundle` asserts non-vacuity — "no sidecar manifests scanned". Every survivor is maintainer-committed and has no sidecar, so that guard fired on a *correct* corpus. It now skips when the corpus holds no bundle-declaring sidecars, and still hard-fails in the case it was written for: sidecars present but none scanned. It re-arms the moment any sidecar lands, which is every submission path.

## Verification

- `validate_corpus.py` → 3/3 cohorts meet the `>=3`-platform floor
- `make pr-preflight` → 28487 passed, 20 skipped
- `explorer_publish.py build` + `results_explorer_snapshot_invariants.py` → pass at 9 results
- `npm run typecheck` and `npm run build` → pass
- Rendered the built site headlessly: 5 ranked platforms across 3 rankings, no console errors, absent coverage shown as "No run" rather than faked, and the Spark SF10 result correctly marked "Excluded" for its one failed query.

## Follow-up

Develop-side deletions do not propagate to `published-results` — that mirror applies a union overlay by design. #1940 is the deliberate counterpart; merge this one first so the two converge.

Nothing prevents this from recurring. Surfacing run age in the Explorer and scheduling a periodic seed-lane refresh would make corpus staleness visible instead of something found by reading timestamps. Not in scope here.


joeharris76:
## `Public-site visual regression` fails — intended, and not merge-blocking

The 12 changed captures are exactly the corpus-driven routes, at all four viewports:

```
/results/            @390 @768 @1280 @1600
/results/benchmarks/ @390 @768 @1280 @1600
/results/platforms/  @390 @768 @1280 @1600
```

Going from 138 bundles across 34 cohorts to 9 across 3 changes what those three pages render. Nothing else in the matrix moved — the landing, docs, and blog routes are unchanged.

**Not a required check.** The `develop-squash-only` ruleset requires `ci-required-result`, `Results Explorer browser gate`, and `ruleset-drift`. All three pass.

**No PR-side fix exists, by policy.** Per `_project/audits/public-site-visual-baseline-policy-2026-08-15.md`, raw screenshots are never committed, and PR diagnostic artifacts "must not become the next baseline without the protected-develop refresh path". Baseline refresh happens only from a successful protected `develop` run after review. So the sequence is: reviewer confirms this diff is intended → merge → the protected develop push run publishes the new baseline.

Worth an actual look at the diff artifacts before approving, since confirming intent is the only gate here. What you should see on `/results/`: 5 ranked platforms across 3 rankings, `No run` cells where a platform has no evidence in a cohort, and `Excluded` for Spark at TPC-DS SF10 (one failed query). I verified that locally against the built site before pushing.

PR #1909, opened 2026-08-25
fix(results): normalize unbacked corpus validation claims

joeharris76:

PR #2001, opened 2026-09-01
docs(adr): record corpus validation-status admission policy

joeharris76:
The not_run entry in PUBLIC_MIRROR_ALLOWED_VALIDATION_STATUSES reads as a
loosening but is a truthfulness tightening (ca08895fd, #1909): it replaced
false passed/partial claims on 52 unvalidated legacy bundles with an honest
label, not the reverse. That reasoning existed only in a code comment and
CORPUS_NOTES.md. Capture it durably before it gets re-litigated from the
diff alone, including the deferred uncertain-admission question as an open
question rather than a decision.


joeharris76:
@codex review

PR #2002, opened 2026-09-01
fix(explorer): show validation badge for non-clean unranked results

joeharris76:
The MetaLeaderboard cell badges (Trust/Funding/Validation) were gated on
cellState.kind === "ranked". Non-clean validation statuses are never
ranked (is_ranking_eligible / ranking_exclusion_reason ==
validation_not_clean), so the ValidationBadge could never fire for the
exact results it exists to flag - including the corpus's 41 not_run
DataFrame-mode bundles (Pandas/Polars/PySpark).

Extend the validation badge to unranked cells with published evidence,
keeping Trust/Funding scoped to ranked cells as before. Also give
"not_run" a warning tone instead of falling through validationTone()'s
neutral default, since "never validated" should not read as the least
alarming status.

### DIFF
--- a/results-explorer/package-lock.json
+++ b/results-explorer/package-lock.json
@@ -2383,9 +2383,9 @@
       }
     },
     "node_modules/baseline-browser-mapping": {
-      "version": "2.10.13",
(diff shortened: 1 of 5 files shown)


joeharris76:
@codex review

PR #2003, opened 2026-09-01
fix(tests): make nightly Windows jobs pass on windows-latest

joeharris76:
Three independent Windows-only failures have kept Nightly Validation red on
both windows-latest/3.10 and 3.12. They are unrelated to each other and are
not regressions of the 2026-08-23..26 nightly batch.

os.fchmod is POSIX-only, so ResultExporter._write_file raised AttributeError
on Windows for every existing destination file. Guard the call and document
that Windows has no equivalent POSIX mode bits to preserve, so the write
proceeds with the new file's default mode.

test_snapshot_file_persistence compared a forward-slash literal against
str(Path), which uses backslashes on Windows; compare against as_posix()
instead of skipping, so the assertion keeps its coverage on every platform.

test_profile_never_turns_architecture_into_cpu_model_without_psutil patched
platform.machine but not platform.processor, so on a real Windows machine it
read the genuine CPU string instead of the intended empty value. It passed
only where the real processor() happened to be empty.

### DIFF
--- a/benchbox/core/results/exporter.py
+++ b/benchbox/core/results/exporter.py
@@ -180,7 +180,11 @@ class ResultExporter:
 
             try:
                 existing_mode = stat.S_IMODE(destination.stat().st_mode) if destination.exists() else None
-                if existing_mode is not None:
+                # os.fchmod is POSIX-only (no Windows implementation); Windows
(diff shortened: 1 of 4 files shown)


joeharris76:
@codex review

PR #2007, opened 2026-09-01
fix(explorer): surface non-clean validation status in default matrix/index view

joeharris76:
The benchmark/platform matrix (QueryHeatmap) hid a non-clean validation
status (e.g. the corpus's 41 not_run DataFrame-mode bundles) behind a
collapsed "Receipt and metadata" disclosure, leaving only a tiny "*"
marker visible by default - a skimming reader saw ordinary heat-mapped
numbers indistinguishable from a validated row. The platform index
(PlatformIndex) carried no validation signal at all in its default
columns; the badge only appeared in the horizontally-scrolled "Source"
column.

Surface a ValidationBadge for any non-"passed" validation_status outside
the disclosure (matrix) and next to the run label (platform index), so
the state is visible without expanding anything or scrolling. Matrix:
move the badge out of <details> instead of duplicating it there. Index:
add it beside the existing "Run" column; the Source column keeps its
own copy as the complete provenance cluster after scroll.

The "not literally passed" predicate mirrors MetaLeaderboard.tsx's
existing gating condition (fix/explorer-validation-badge-gating,
unmerged) and now lives once in lib/displayLabels.ts as
isValidationNotClean() rather than being re-derived per surface. It is
a placeholder for the shared status->reader-facing-label mapping being
introduced on fix/explorer-compare-validation-disclosure; reconcile
once that lands.

### DIFF
--- a/results-explorer/package-lock.json
+++ b/results-explorer/package-lock.json
@@ -2383,9 +2383,9 @@
       }
     },
     "node_modules/baseline-browser-mapping": {
-      "version": "2.10.13",
-      "resolved": "https://registry.npmjs.org/baseline-browser-mapping/-/baseline-browser-mapping-2.10.13.tgz",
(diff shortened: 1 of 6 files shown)


joeharris76:
@codex review

PR #2026, opened 2026-09-03
docs(agent): state the disclosure condition in plain language

joeharris76:
`docs/agent/attribution-surfaces.md` described when an agent should disclose in
the text by calling that condition "load-bearing". The metaphor names no
condition, and the two examples that follow it already state one.

This replaces the phrase with the condition itself — whether the disclosure
changes what a reader would do — and rewraps the paragraph to the file's
80-column width. Wording only; the rule's meaning is unchanged.


joeharris76:
@codex review

PR #2029, opened 2026-09-03
docs(agent): word CI-status rule as scheduled checks, not a polling ban

joeharris76:
One-line fix: agents read 'Do not poll CI: pending is a valid terminal state' as a hard stop on all CI status reads, stalling merge/deploy monitors. The intended rule was bounded scheduled checks (sleep/cron between reads), never spin loops. No policy change, just wording.


joeharris76:
@codex review

PR #2031, opened 2026-09-03
fix(publication): compare canonical DB digest in G2 root-neutrality gate

joeharris76:
Exact results.duckdb bytes can never match across builds: the explorer
pipeline stamps every build with a wall-clock generated_at and float
aggregates such as geomean_ms drift 1 ULP across runners (observed live:
same-input rebuilds 60s apart hash differently with identical logical
content). The gate now compares canonical content digests via
scripts/publication/compare_db_digest.py, still failing closed on
missing/unreadable input, live-download failure, live movement, or any
added/removed table, column, row, rank, or material value change.


joeharris76:
@codex review

PR #2032, opened 2026-09-03
Retire _project/scripts/todo shim, adopt todo-db-mcp

joeharris76:
Retires the `_project/scripts/todo` bash shim and its compatibility adapter now that the tracker interface is the vendored `todo-db` 0.6.1 wheel plus the `todo-db-mcp` server. Unblocks and closes the retire-todo-shim-adopt-todo-db-mcp tracker item.

## Commits (10, in order)

1. `9d00b006f` vendor 0.6.0 wheel + `mcp` extra
2. `f23297855` w0 review fixes (mcp-extra assertion, decision record)
3. `5ea23a85b` adopt `todo` skill v1.0.0 (skill-sync @ 20d53d9)
4. `8172d6c` register `todo-db-mcp` (`.mcp.json`, credential docs)
5. `9204a7a65` render export views from the floor-CLI envelope (new `todo_db_export_views.py`)
6. `8f0e831f3` read tracker state from the floor-CLI export envelope (reconciliation + schema gate off the shim)
7. `393431a21` remove the bash shim and the compatibility adapter (-2,922 lines)
8. `0cb49f4ea` drop the inert `wrapper` key from the tracker config
9. `37401379` move the tracker docs to the MCP and floor-CLI model
10. `efb4d5e3` vendor 0.6.1 wheel and re-pin skill-sync to d22ea7f (PR #21 MCP additions + PR #71 prioritize fix)

## Verification

- Fast suite: 29,250 passed, 19 skipped (final branch state, local).
- Instruction-parity audit + docs surfaces: green.
- Schema contract gate: OK. 0.6.1 keeps SCHEMA_VERSION 7 with byte-identical migrations, so no hosted migration.
- Wheel SHA-256 matched the v0.6.1 release SHA256SUMS; MCP stdio handshake verified in commit 4.
- Pre-commit hooks (audit, markdownlint, ruff, codespell) pass on the new commits.

## Reviewer attention: CI-only coverage

The export-workflow repoint and the reconciliation rewrite exercise hosted credentials, so they only run green in CI. Please confirm the `todo-db-export` and `pr` workflow runs on this branch before merging.

PR #2034, opened 2026-09-04
Repoint publication reconciliation to the floor-CLI export envelope

joeharris76:
First half of retiring `_project/scripts/todo shim: the publication reconciliation gate reads live tracker state from one atomic `todo-db export` envelope instead of shelling the shim's removed-in-spirit `list`/`show` verbs. Fail-closed semantics unchanged.

Ordering matters: this lands while the shim still exists because the credentialed CI step runs the base-branch script whenever this file changes. Verified the new invocation (`export --output`, `--project-id`/`--repository`, `TODO_DB_RO_AUTH_TOKEN`, `TODO_DB_AUTH_CONTRACT=v2`) against the currently vendored 0.4.3 wheel's CLI, whose envelope shape matches 0.6.x. Unit tests (32) and ruff pass locally. The shim's deletion follows in #2032, which rebases onto this.


joeharris76:
@codex review

PR #2035, opened 2026-09-04
fix(publication): reuse probe deploy run in soak conclude job

joeharris76:
Dispatch titles are the workflow name, never the generation, so the
conclude job's title search resolved null and the artifact fetch 404'd
(observed on the first gen1 conclude dispatch). The probe now emits the
resolved deploy run id; conclude consumes it fail-closed. Explicit
generation input is verified against the downloaded record instead of
searched by title.


joeharris76:
@codex review

PR #2067, opened 2026-09-06
fix(explorer): render charts inside their own drawing at every width

joeharris76:
## What was wrong

Three independent defects, each of which left a chart that still looked finished.

**1. SVG attribute names were being dropped.** SVG attribute names are case-sensitive, and Preact forwards a prop it does not recognise to `setAttribute` under the name it was given. Nine chart components used the camelCase spellings, so `textAnchor`, `strokeWidth`, `strokeDasharray` and `fillOpacity` arrived in the DOM as attributes the renderer has no rule for, and were ignored. In the browser:

```html
<text x="146.5" y="30" textAnchor="middle" ...>Geomean query time (log scale) - lower is better</text>
```

That label is not centred. Nor is any other. Every label meant to be right-aligned against its gutter started at the gutter edge and ran across the bars; every emphasised stroke drew at 1px (`getComputedStyle` → `stroke-width: 1px` against an authored `1.5`); every dashed stroke drew solid (`stroke-dasharray: none`) — including the histogram marker whose stated purpose is to make a query that did not run look different from one that ran very fast, and the box plot's whiskers and its 0.18-opacity IQR box, which rendered fully opaque over them.

**2. Eight charts drew outside their own box.** They authored themselves at `Math.max(containerWidth, 400)` with no `viewBox`. `.bb-chart-svg { max-width: 100%; height: auto }` then reduced the box without moving the coordinate system. Measured on the live site at a 390px viewport, chart column 293px:

| | |
|---|---|
| SVG attributes | `width="400" height="472"`, no viewBox |
| Rendered box | 293 × 345.7 |
| Never painted | SQLite, Spark, PySpark; every x-axis tick; the caption naming the log scale |
| `parent.scrollWidth` vs `clientWidth` | 293 vs 293 |

That last row is the trap: the wrapping `.overflow-x-auto` cannot scroll to the missing quarter, because the box now fits its parent.

**3. The CDF legend ran off the canvas.** Entries are placed at `36 + 130 × i` on one row with no wrapping. On TPC-H SF 1 there are twelve series, so the row is 1,496 units wide inside a 1,166-unit drawing. Read from the live DOM, three entries sit at x = 1206, 1336 and 1466 — SQLite, Spark and PySpark, the three slowest runs and the curves furthest right.

## What changed

Charts now draw in a coordinate system as wide as their measured container and publish it as `width="100%"` plus a matching `viewBox` (`src/lib/chartFrame.ts`). Because the viewBox width and the rendered width agree, the user-unit scale is exactly 1 and label size is screen-fixed by construction — not by CSS, which would not have helped: `font-size` inside an SVG is measured in user units whether it arrives as an attribute or a rule.

Below 480px a bar row reflows instead of holding a gutter it cannot afford: the label moves onto its own line above the bar, the value keeps the opposite end of that line where no bar length can displace it, and the bar spans the full width.

The CDF legend wraps in HTML outside the SVG, matching the pattern already used by `StackedPhase` and `TimeSeries`.

Smaller fixes in the same pass:

- The per-query histogram split panels on query count alone, so a wide cohort clamped each bar to a floor wider than the group holding it and consecutive groups painted over one another. Panels now account for cohort size.
- Axis labels on the outermost tick tuck inside the drawing rather than hanging half outside it.
- The heatmap announced a cell as `2.00× fastest in column` when the ratio is this run over the column minimum — the cell is 2× *slower*. Corrected in both the grid's accessible names and the mobile cards.
- `QueryTimingChart` had no caller; removed.

Separately, the panel headed **Charts** opened on the sparkline table — an HTML metrics table — so a reader who never touched the controls saw no chart at all. A cohort summary now opens on the performance bar; the table stays available in the Overview group. The open chart moves into a `chart` URL parameter so a reader who finds the view that answers their question can send that view; an id that does not apply to the cohort is canonicalised and the parameter dropped.

## Verification

- Unit suite: 93 files, 1,310 passed.
- Chromium e2e for `responsive`, `benchmark-index`, `result-detail`: 38 passed.
- Walked every chart in every question group against a local build at 375px and at 1280px: viewBox width equals container width in all cases, zero overlapping label pairs, nothing outside the drawing on either axis.

## New coverage

The document-overflow audit in `e2e/responsive.spec.ts` exempts anything inside `svg[role='img']`, which is why none of this was caught. Added:

- `e2e/responsive.spec.ts` — each chart's drawing width must match its rendered width and no label may fall outside it. Confirmed this fails against the previous code (`Geomean performance comparison: no viewBox`).
- `src/components/__tests__/chartResponsiveContract.test.tsx` — the sizing contract per chart at a phone and a desktop column.
- `src/__tests__/svgAttributeHygiene.test.ts` — the camelCase spellings cannot return.
- `src/lib/__tests__/chartFrame.test.ts` — frame and row geometry.
- Histogram group non-overlap and CDF legend completeness, both confirmed to fail against the previous code.

`GroupedQueryChart.test.tsx` accepted either `stroke-dasharray` or `strokeDasharray`; it now requires the name that works.

## Deliberately not in this change

An earlier draft proposed opening the route on three stacked charts with a grid of live chart previews below. An adversarial review argued that down, and I agree: `BenchmarkIndex` already renders the query heatmap as its primary matrix and excludes it from the panel, so the default set would duplicate it; live previews instantiate the full chart rather than a cheap image; and a thumbnail that omits its exclusion summary asserts "these are the compared platforms" when the claim is "these are the rows eligible under this chart's evidence policy". The panel keeps one fully disclosed active chart.

Known and not addressed here, all worth their own change: the mobile heatmap shows only three computed outliers per platform with no way to reach the rest, so the available evidence changes with viewport; `CostScatter` has no collision handling, so equal points occlude silently; `TimeSeries` spaces dates ordinally, so a one-day gap and a one-year gap look alike; `MultiRunHeatmap` ignores its own default render limit and its table has no accessible name; `MetaLeaderboard` publishes `aria-colindex`/`aria-rowindex` without offsetting for the platform column and header row; and most chart marks carry their values only in `<title>` on unfocusable elements, which is not a dependable screen-reader interface.

### DIFF
--- a/results-explorer/e2e/responsive.spec.ts
+++ b/results-explorer/e2e/responsive.spec.ts
@@ -256,41 +256,78 @@ test.describe("responsive explorer assertions", () => {
       await page.goto("/results/tpch/");
       await waitForDataLoaded(page, /Charts/);
 
-      const offenders = await page.evaluate(() => {
-        const problems: string[] = [];
-        for (const svg of Array.from(document.querySelectorAll("svg[role='img']"))) {
-          const box = svg.getBoundingClientRect();
-          if (box.width === 0) continue;
+      // Walk every question group and every chart within it. Checking only the
+      // default view leaves most of the chart set unmeasured, and the default
+      // is the one chart least likely to be wrong.
+      const groups = page.getByRole("tablist", { name: "Chart question groups" }).getByRole("tab");
+      const groupCount = await groups.count();
+      // If this ever reads zero the walk below is vacuous and the test would
+      // pass without measuring anything.
+      expect(groupCount, "chart question groups").toBeGreaterThan(1);
+      let chartsVisited = 0;
 
-          const viewBox = svg.getAttribute("viewBox");
-          if (viewBox === null) {
-            problems.push(`${svg.getAttribute("aria-label") ?? "chart"}: no viewBox`);
-            continue;
+      for (let g = 0; g < groupCount; g += 1) {
+        await groups.nth(g).click();
+        await page.waitForTimeout(150);
+        const chartButtons = page.locator('[aria-label$="charts"] button');
+        const chartCount = await chartButtons.count();
+
+        for (let c = 0; c < Math.max(chartCount, 1); c += 1) {
+          if (chartCount > 0) {
+            await chartButtons.nth(c).click();
+            await page.waitForTimeout(150);
           }
+          chartsVisited += 1;
 
-          // Drawing width must match rendered width, or the coordinate system
-          // and the box disagree and marks fall outside the paint area.
-          const drawWidth = Number(viewBox.split(/\s+/)[2]);
-          if (Math.abs(drawWidth - box.width) > 1.5) {
-            problems.push(
-              `${svg.getAttribute("aria-label") ?? "chart"}: draws ${drawWidth} in a ${box.width.toFixed(0)}px box`,
-            );
-          }
+          const offenders = await page.evaluate(() => {
+            const problems: string[] = [];
+            for (const svg of Array.from(
+              document.querySelectorAll("[data-chart-container] svg[role='img']"),
+            )) {
+              const box = svg.getBoundingClientRect();
+              if (box.width === 0) continue;
 
-          for (const text of Array.from(svg.querySelectorAll("text"))) {
-            const textBox = text.getBoundingClientRect();
-            if (textBox.width === 0) continue;
-            if (textBox.right > box.right + 1 || textBox.bottom > box.bottom + 1) {
-              problems.push(
-                `${svg.getAttribute("aria-label") ?? "chart"}: "${(text.textContent ?? "").slice(0, 24)}" falls outside the drawing`,
-              );
+              const viewBox = svg.getAttribute("viewBox");
+              if (viewBox === null) {
+                problems.push(`${svg.getAttribute("aria-label") ?? "chart"}: no viewBox`);
+                continue;
+              }
+
+              // A drawing narrower than its box is scaled UP, which magnifies
+              // every coordinate including the gaps bars were spaced by.
+              const drawWidth = Number(viewBox.split(/\s+/)[2]);
+              if (drawWidth < box.width - 1.5) {
+                problems.push(
+                  `${svg.getAttribute("aria-label") ?? "chart"}: draws ${drawWidth} into a ${box.width.toFixed(0)}px box`,
+                );
+              }
+
+              for (const mark of Array.from(svg.querySelectorAll("text, rect, circle, path"))) {
+                const markBox = mark.getBoundingClientRect();
+                if (markBox.width === 0 && markBox.height === 0) continue;
+                if (
+                  markBox.right > box.right + 1 ||
+                  markBox.left < box.left - 1 ||
+                  markBox.bottom > box.bottom + 1
+                ) {
+                  const what =
+                    mark.tagName === "text"
+                      ? `"${(mark.textContent ?? "").slice(0, 24)}"`
+                      : mark.tagName;
+                  problems.push(
+                    `${svg.getAttribute("aria-label") ?? "chart"}: ${what} falls outside the drawing`,
+                  );
+                }
+              }
             }
-          }
-        }
-        return problems;
-      });
+            return problems;
+          });
 
-      expect(offenders, offenders.join("\n")).toEqual([]);
+          expect(offenders, offenders.join("\n")).toEqual([]);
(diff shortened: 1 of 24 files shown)


joeharris76:
## Adversarial review of the implementation, and what changed

Ran an adversarial review over the diff itself rather than the plan. Six required findings; five accepted and fixed in `444bc272`, one rebutted.

### Accepted

**Value labels outside the drawing at any width.** The review's sharpest finding. A label that trails the end of a bar was placed just past it, so whenever a bar reaches the edge of the plot the label is drawn outside the viewBox — and a clamped maximum *always* reaches the edge. `NormalizedSpeedupChart` lost the label on any query at or beyond its clamp, `DivergingBarChart` on any delta at or beyond ±100%, and `PercentileLadder` lost its p50 value whenever p50 sat near p99. New `edgeSafeValueLabel` turns the label back on itself so it reads from inside the bar.

**`GroupedQueryChart` magnified its own drawing.** Its viewBox was sized to the container while its CSS held the box open to `groupCount * 60`px for scrolling, so the drawing was stretched by the ratio between them — twentyfold at a hundred groups — after computing bar widths against a width it never had. Bars collided at their minimum, then got blown up along with the gaps. The viewBox now matches the scroll minimum. This was a pre-existing defect the first commit did not touch.

**The histogram's minimum bar width was treated as a guarantee.** A cohort too large for one bar per platform in a single group still overran it. The minimum is now a preference: past that point the bars go thin rather than the groups colliding. Query labels narrower than a label are thinned rather than overprinted.

**The compact axis caption did not fit.** I had anchored it left, which was not enough — at the 240-unit floor it still ran to ~288. A compact drawing now names the scale and direction and leaves the metric to the heading and accessible name, which already carry it. The review was right and my first fix was insufficient; the new 240px test case is what proved it.

**URL state lifecycle.** `useUrlState` passed `null` as replaceState's state argument, discarding whatever the entry held for the router or any other consumer — now preserved. And router navigation between two routes that both host the panel uses pushState, which fires no popstate, so the panel kept the reader's chart while the destination URL said nothing about it; the parameter is now brought back into step. Writing that fix surfaced a bug in itself — the sync effect re-asserted an id the canonicalisation had just rejected — caught immediately by the existing canonicalisation test, and now guarded on applicability.

### Rebutted

**"No changed test checks the CDF legend."** There is one: `charts.smoke.test.tsx`, *keys every series, including the slowest, at any width* — twelve series, twelve `<li>`, asserting the last is present. That is the defect the HTML legend fixes.

**"`axisLabelAnchor` cannot guarantee labels fit."** For the failure it addresses — a label hanging off an edge — anchoring `end` at the right edge fits by construction, because the label extends inward. The residual risk is interior labels colliding with each other, which is a different failure mode and not what the helper claims.

### Test criticisms, accepted

The review was right that several tests asserted less than they appeared to. Fixed:

- `chartResponsiveContract` covered six charts; it now covers ten, including the two whose edge bug this commit fixes and the two (`CostScatter`, `TimeSeries`) that were omitted.
- It read only rects, line endpoints and text x — so the CDF's plotted curves, its entire data layer, went unexamined, and no vertical extent was checked at all. It now reads path and polyline geometry, circles, and both axes.
- Requiring viewBox width to equal rendered width rejected charts that legitimately declare their own floor. Each chart now states its floor and the assertion respects it.
- Removing `axisLabelAnchor` from every call site used to leave the suite green. It now fails.

I verified each of these by reverting the fix and confirming the test goes red, rather than assuming.

The stated counts in the hygiene test were wrong — ten files and 68 occurrences, not nine and 70 — and its limits are now written down: it is a lexical scan, so it cannot see a spread, a computed key, or a runtime-assembled name.

### The CI failure

The review disproved my hypothesis that the `chart` parameter was damaging Compare's `ids`, tracing every step from the router-owned URL through the notices to the panel, which does not mount until loading completes. The evidence agrees: the full chromium suite passes locally, the failure screenshot shows the page rendered correctly with all three notices present, and a re-run of the unchanged commit came back green. It was a flake at a data-bound wait, not a regression.

While reading that code the review also found a pre-existing race in Compare's asynchronous ID canonicaliser — it captures results, awaits `toShortIds`, then writes to whatever route is current, so navigating away mid-await can resurrect old `ids` on another route. Not introduced here and not fixed here.

### Still not addressed

Everything in the original "Deliberately not in this change" section stands, plus: below `CHART_MIN_WIDTH` the drawing does scale, so text shrinks with it — a deliberate trade, now stated accurately in `chartFrame.ts` rather than claimed as 1:1 at every width.

PR #2111, opened 2026-09-13
fix/sqlglot upstream repro refresh

joeharris76:
## Summary

Add a standalone SQLite execution reproducer for date extraction, with tests
that distinguish incorrect results from execution errors. Record timestamp-cast
and integer-division counterexamples so a simple STRFTIME replacement is not
mistaken for a semantics-preserving translation.

Correct the interval result representation and QuestDB CTE alias example, and
require native SingleStore evidence before making a MySQL percentile claim.
These changes affect contribution preparation only; production translation and
dependencies are unchanged.

The extraction evidence has been added to the existing upstream discussion:
https://github.com/tobymao/sqlglot/issues/2592#issuecomment-5653487646

## Validation

- Run `uv run -- python -m pytest -q tests/unit/scripts/test_sqlite_extract_repro.py`.
- Run the standalone reproducer with SQLGlot 30.6.0 and 30.18.0. Exit 1 is the
  expected diagnostic result while the translation failures remain.
- Run repository `make pr-preflight`.

The reproducer executes SQLite, not PostgreSQL. Its explicit expected results
use DATE columns represented as ISO date TEXT; it does not claim universal
equivalence or authorize workaround retirement.


joeharris76:
Upstream submissions:

- SQLite binary DATE + INTERVAL: https://github.com/tobymao/sqlglot/issues/8347
- QuestDB dialect support direction: https://github.com/tobymao/sqlglot/issues/8348
- SQLite extraction follow-up: https://github.com/tobymao/sqlglot/issues/2592#issuecomment-5653487646

The interval report demonstrates a DuckDB TIMESTAMP result and SQLite execution
failure on SQLGlot 30.18.0 and main at
5cfb5997a99010940138670adf3d6b34ac5a0a08. It makes no general claim that DATE or
DATETIME modifier rewrites preserve every interval expression.

Discussions is disabled upstream, so the QuestDB issue asks for the preferred
support route without repeating unverified engine restrictions or committing
to a maintained dialect.

The percentile draft is not submitted: write="singlestore" removes the reported
CASE decorator and emits ORDER BY x NULLS LAST. No local SingleStore listener
was found on the standard port; that remaining form has not been engine-tested.
The earlier MySQL-wide claim is not established.

These links supersede the preparation snapshot's unfiled status for interval
and QuestDB. No translator fixes or workaround removals have been submitted.

joeharris76:
SingleStore validation is now complete for the bounded percentile reproduction:
https://github.com/tobymao/sqlglot/issues/8349

The local stack is `docker/singlestore/docker-compose.yml` on port 13306.
The earlier search of standard ports missed that configured environment.
With Mocker 0.7.2, the first image fetch selected ARM64 despite the compose
platform setting. An explicit AMD64 pull and CONTAINER_DEFAULT_PLATFORM=linux/amd64
resolved that step. The 1 GiB VM then exited with a SingleStore startup assertion;
a temporary 4 GiB override started SingleStore 9.1.1 with Rosetta enabled.
Compose's Started output alone did not establish database readiness.

An initial fixture used CAST AS DOUBLE, which SingleStore rejected; those
results were not counted as percentile evidence. The corrected DECIMAL fixtures
and portable empty relation produced 25/25 matches for the original SQL,
0/25 for the MySQL-targeted SQL (parse errors), and 23/25 for native SingleStore
output. The two native mismatches involved mixed NULLs at percentiles 0.75 and 1.

The isolated three-value reproduction removes casts: PERCENTILE_CONT(1) over
[NULL, 1, 2] returns 2 without explicit NULL ordering, 2 with NULLS FIRST, and
NULL with NULLS LAST. SQLGlot 30.18.0 and inspected main emit NULLS LAST.
The upstream report leaves engine-versus-generator ownership open rather than
claiming a globally safe NULL-ordering removal.

Only this test stack and its temporary volume were removed. No production
translation, dependencies, or compose definitions were changed.

joeharris76:
## SQL transpiler evaluation status — 2026-09-13

Production translation and dependencies remain unchanged. Further SQLGlot submissions are paused while we assess whether BenchBox-local extensions, a private SQLGlot fork, or a Polyglot extension/fork is the lowest-maintenance choice.

- SQLGlot SQLite issue #8347 and SingleStore issue #8349 were closed as not planned without explanatory comments. QuestDB #8348 was explicitly declined. The SingleStore fix PR https://github.com/tobymao/sqlglot/pull/8352 remains open and draft, with no review at this check; not all submissions were rejected.
- Prepared SQLGlot branches remain in joeharris76/sqlglot: fix/sqlite-date-interval-arithmetic, fix/postgres-extract-return-types, and dependent fix/sqlite-date-extraction. Prior local checks passed; their status does not imply upstream acceptance or BenchBox integration. Native SingleStore translation also differs from BenchBox's current mysql target.
- Fresh Polyglot 0.9.2 versus SQLGlot 30.18.0 probes show mixed results. Polyglot handles SQLite DATE extraction with fractional division, but changes PostgreSQL integer division: SELECT 2020 / 3 > 673 becomes true instead of false. This expectation follows PostgreSQL documentation; generated SQLite was executed, not a new PostgreSQL server.
- Both libraries still emit invalid SQLite SQL for binary DATE + INTERVAL. Polyglot's ORDER BY ALL handling causes a DuckDB column error and can execute with incorrect ordering in SQLite. Its Python transpile API rejects BenchBox's identify=True option.
- Polyglot has Rust custom-dialect hooks worth testing before assuming a core fork is necessary. Existing SQLGlot AST consumers and native-wheel ownership make this more than a central-call replacement. Earlier performance logs favor Polyglot, but were not rebenchmarked and do not establish migration safety.

The following TODOs preserve the remaining work:

1. `sql-transpiler-comparison-evidence` — recover reproducible comparison artifacts and refresh known-good/problem execution and performance results.
2. `polyglot-source-aware-division-spike` — test one bounded semantic patch through custom hooks first; compare maintenance effort with the prepared SQLGlot patches.
3. `polyglot-benchbox-integration-assessment` — assess quoting, ordering, Python/AST compatibility, packaging and release ownership.
4. `sql-transpiler-maintenance-decision` — depends on the three investigations; recommend retention or migration and reconcile the existing SQLGlot ADR.

No production migration, workaround retirement, new hosted fork, additional upstream submission, or renewed model-training effort is authorized by these research tasks. A production implementation requires a separate decision and approval.

PR #2141, opened 2026-09-15
chore(deps): upgrade datafusion 53.0.0 to 54.0.0

joeharris76:
Adapt to datafusion 54 API changes: pass Python literals to
date_part/date_trunc/extract part arguments (Expr form deprecated)
and compare the string-typed fixture column to a string literal
(mixed string/int comparison is now a cast error).


joeharris76:
Follow-up landed as #2149: all six datafusion declarations move to `>=54.0.0`, matching the locked 54.0.0 (floor-raise option, no version-gated shim). Lock diff is specifier metadata only; preflight green.

PR #2149, opened 2026-09-15
fix(deps): raise datafusion floor to 54 for the str-arg function API

joeharris76:
# fix(deps): raise datafusion floor to 54 for the str-arg function API

Follow-up to a P1 review thread on #2141, which merged before the fix
could land on it.

## Problem

#2141 moved `date_part` / `date_trunc` / `extract` to the plain-string
argument form, which DataFusion only accepts on 54+. The declared
`datafusion>=50.1.0` (and dev `>=40.0.0`) floors still admitted 53,
where those calls raise `AttributeError: 'str' object has no attribute
'expr'`.

## Change

Took the reviewer's floor-raise option over a version-gated shim: all
six declarations move to `datafusion>=54.0.0`, matching the locked
54.0.0.

- `uv.lock`: specifier metadata only — zero version changes.
- `docs/development/dependency-audit-raw.md`: regenerated with the
  repo's own script.

## Verification

- `make pr-preflight` green on this head.
- `uv lock --check` passes.


joeharris76:
Follow-up complete: registry requirement `34.0.0→54.0.0`, checker floor `40.0.0→54.0.0`, and below-minimum installs now report unavailable with an upgrade error (plus regression tests). Full preflight green on this head.

PR #2140, opened 2026-09-15
chore(deps): upgrade database/cloud SDKs to latest compatible releases

joeharris76:
- pyathena 3.30.1 -> 3.36.0
- boto3 1.42.83 -> 1.43.94 (botocore, s3transfer bumped transitively)
- redshift-connector 2.1.13 -> 2.1.16
- singlestoredb 1.16.9 -> 1.17.3
- firebolt-sdk 1.18.5 -> 1.18.6
- google-cloud-bigquery 3.41.0 -> 3.45.1 (lxml, aiorwlock bumped transitively)
- snowflake-connector-python 4.4.0 -> 4.7.3
- presto-python-client and azure-identity already current; skipped

All floors in pyproject.toml remain satisfied; Python >=3.11 floor
unchanged; numpy/scipy pins untouched.

PR #2150, opened 2026-09-15
fix(ci): drop obsolete lxml PYSEC-2026-87 audit exception

joeharris76:
# fix(ci): drop obsolete lxml PYSEC-2026-87 audit exception

Follow-up to a P2 review thread on #2140, which merged before the fix
could land on it.

## Problem

The SDK upgrades moved lxml past 6.1.0 (lock resolves 6.1.3), but
`pr.yml`, `test.yml`, and `nightly.yml` still passed
`--ignore-vuln PYSEC-2026-87`. A later resolution reintroducing a
vulnerable lxml would have been silently accepted by every configured
audit.

## Change

- Removed the ID (and its stale redshift-connector comment) from all
  three ignore lists; `test.yml` / `nightly.yml` now pass an empty
  list, which the guarded workflow step treats as "no ignores".
- Kept the unrelated sqlparse exceptions in `pr.yml` untouched.

## Verification

- `make pr-preflight` green on this head.
- Lock already resolves redshift-connector 2.1.16 (constraint lifted)
  and lxml 6.1.3.


joeharris76:
Complete: readiness-queue entry removed alongside the ignore-list removal (tracker item close-out still pending — the todo-db state branch is unreachable from this host; will retry). Full preflight green on this head.

PR #2159, opened 2026-09-17
fix/datavault cross surface burn down

joeharris76:
- **fix(datavault): burn down cross-surface query execution divergences**
- **fix(datavault): address independent review findings**
- **fix(datavault): close delta-review findings on grouped aggregates**
- **fix(datavault): mark PySpark idiom tests slow per marker policy**


joeharris76:
@codex review

PR #2166, opened 2026-09-17
test(qpc14): pin fingerprint_version guard on legacy upgrade boundary

joeharris76:
Extend the v1-then-v2 boundary test to three post-boundary runs plus a
v1 rollback so the no-flap assertion exercises the version guard (two
runs short-circuit the check) and version history covers downgrades.


joeharris76:
@codex review

PR #2167, opened 2026-09-17
fix(spark): port AQE-toggle fixes to sibling Spark-family adapters

joeharris76:
Hoist AQE/CBO keys and the resolve-then-apply runtime helper into
_spark_helpers, set AQE keys explicitly in both directions at session
build for lakesail, velox, pyspark session manager, and lakesail_df,
honor the adapter toggle with spark_config precedence at runtime, add
the lakesail adaptive_enabled option, and merge user spark_config last
on Synapse/Fabric Livy sessions. Cloud toggle follow-up tracked
separately.


joeharris76:
@codex review

PR #2168, opened 2026-09-17
fix(dataframe): record dask spill directory on effective consumption

joeharris76:
Record a user-configured spill directory when the cluster actually
consumes it (local_directory), even when spilling itself came from
envelope defaults rather than the tuning flag — while never claiming
auto-created temp dirs, the spill_to_disk default, or a bare
constructor directory with no tuning config.


joeharris76:
@codex review

PR #2170, opened 2026-09-17
fix(starrocks): render distribute_by as SQL with dialect clause ordering

joeharris76:
Store fully rendered distribution SQL in TuningClauses.distribute_by
(never a bare column), order DISTRIBUTED BY after PARTITION BY for
StarRocks and Doris (DUPLICATE KEY first) via a platform marker, drop
empty markers from serialization, and show distribute_by in the CLI
dry-run summary — so previews and execution DDL agree.


joeharris76:
@codex review

PR #2172, opened 2026-09-17
fix(dataframe): return real list expressions from DataFusion str.split

joeharris76:
Use string_to_array (available since DataFusion v54 bindings) instead
of the split_part proxy, so split results support full list semantics
(get/len/contains/collect). Out-of-range indices now yield NULL,
matching polars and pyspark.


joeharris76:
@codex review

PR #2173, opened 2026-09-17
fix(dataframe): coerce inferred TIME columns to string in Parquet cache

joeharris76:
PyArrow CSV inference types HH:MM:SS fields as time32/time64, which
Parquet stores as TIME(MILLIS/MICROS,false) that Spark rejects
(PARQUET_TYPE_ILLEGAL). Persist them as strings, matching the schema
maps, and bump the cache version so stale TIME caches regenerate.


joeharris76:
@codex review

PR #2175, opened 2026-09-17
feat(pyspark): close read-primitive map/fulltext/window/array gaps

joeharris76:
Add lag/lead/ntile/percent_rank/cume_dist window methods plus ordered
collect (sort/sort_by), Python-type casts, and a registry-derived PySpark
execution sweep. Skip 12 raw-Polars impls pending porting.


joeharris76:
@codex review

PR #2178, opened 2026-09-17
Propagate per-table load timings into benchmark results

joeharris76:
Adapters that already time each table now report those measurements
instead of discarding them: Snowflake records load_start elapsed time
per table, and BigQuery threads per-table elapsed time out of its
cloud-storage and direct load helpers. MySQL-wire adapters add total_ms
alongside the existing duration_seconds field, and Snowpark wraps its
per-table seconds in the shared {total_ms} mapping so every producer
conforms to the result factory contract.

The full benchmark path previously dropped these timings before result
construction, so the tables block never carried load_ms outside
load-only runs. The setup phases now retain the adapter-provided
timings as run-scoped state (reused databases correctly retain none)
and run_enhanced_benchmark forwards them to result construction, where
missing or malformed entries continue to omit load_ms without raising.

Verified: fast unit gate 28348 passed, ruff check and format clean,
duplicate-line delta +0.


joeharris76:
@codex review

PR #2179, opened 2026-09-17
fix/first live cloud benchmark run

joeharris76:
## Problem

Result bundles from Databricks SQL Serverless recorded
`"engine_version": "4.2.0 0000000000000000000000000000000000000000"`.
`SELECT version()` returns the raw Spark engine version plus a 40-zero
commit-hash placeholder, not the Databricks release. A live probe of
`SELECT current_version()` on the warehouse returns the true release struct,
e.g. `dbsql_version='2026.36'` with `u_build_hash`/`r_build_hash`.

## Change

`benchbox/platforms/databricks/adapter.py`
(`get_platform_info()`, `_get_platform_metadata()`, new
`_resolve_databricks_engine_version()` helper):

- Probe `SELECT current_version()` first. Warehouses record `dbsql_version`,
  clusters record `dbr_version`, as `platform_version`/`engine_version` with
  `engine_version_source="current_version"`, plus build hashes.
- Fall back to `SELECT version()` then `SELECT spark_version()` where
  `current_version()` is unavailable, sanitized so the placeholder hash never
  reaches stored metadata. The sanitizer strips the second token only when it
  matches a commit hash (`^[0-9a-fA-F]{7,40}$`), preserving legitimate
  multi-token versions such as `"Runtime 14.3 LTS"`.
- Follow-up hardening: scalar guard in `_first_column` (a bare string must not
  be truncated to its first character), private provenance constants, and
  removal of a dead `spark_version_raw` branch.

On the provenance value: `engine_version_source="current_version"` is
deliberately more specific than `"sql_query"`. The vocabulary in
`benchbox/core/results/models.py` is already open in practice (Athena emits
`"config"`), no consumer branches on these values, and the six remediated
result bundles below already persist `"current_version"`, so renaming would
split history.

## Tests

- 25 new unit tests in
  `tests/unit/platforms/test_databricks_adapter_coverage.py` covering dict and
  connector-`Row` structs, DBR selection, `current_version()` error/`None`
  fallback, `spark_version()` fallback, total failure, and metadata paths;
  updated the `test_get_platform_metadata` fixture for the new probe order.
- Full Databricks suite: 347 passed. `ruff check` and `ruff format --check`
  clean.

## Result bundles (outside this repo)

Six files under `benchmark_runs/results/` and `/tmp/live_cloud_cohorts/`
(`tpch_sf01/sf1/sf10_databricks_sql_20260917_*.json`) now record
`"2026.36"` with `dbsql_version`, `u_build_hash`, `r_build_hash`, and
`engine_version_source="current_version"`; no placeholder hashes remain.


joeharris76:
@codex review

PR #2184, opened 2026-09-18
Fix live plan-capture defects found in adversarial review

joeharris76:
Snowflake get_query_plan re-serializes non-string VARIANT cells with
json.dumps instead of str(), whose repr output the plan parser rejects.
BigQuery live capture tests now apply the benchmark job config so repeated
SELECT 1 executions bypass the result cache and expose full job statistics,
and assert through .get() lookups for clearer failures.


joeharris76:
@codex review

PR #2188, opened 2026-09-18
fix(plan-capture): return None on EXPLAIN failure in mysql-wire, fabric, questdb

joeharris76:
Log a warning and return None instead of an error string so generic capture records explain_failed rather than misclassifying the failure as parse_error.


joeharris76:
@codex review

PR #2189, opened 2026-09-18
fix(snowflake): return join conditions as list for deterministic sorting

joeharris76:
Return stripped condition strings instead of one comma-joined string so the structural signature sorts multi-condition joins deterministically.


joeharris76:
@codex review

PR #2191, opened 2026-09-18
fix(nightly): correct -m filter docs and cover Snowflake execute_query plan attach

joeharris76:
Clarify that the -m marker filter selects within the node-id classes rather than backstopping node-id drift. Add a live execute_query test verifying plan and fingerprint attachment.


joeharris76:
@codex review

PR #2196, opened 2026-09-18
fix/status wiring region guard

joeharris76:
Routes fallback-priced and stale runs into `cost_status: unavailable`.

This merges after #2195: it is stacked on `fix/fail-closed-price-resolution`, so the diff reads clean once that PR lands.

What changes:

- `_normalized_cost_warnings` scans `phase_costs[*].query_costs[*].pricing_details` for the `price_unavailable` marker and emits one warning per fallback table. The existing machinery then flips `cost_status` to `unavailable`, and `NormalizedCost` drops the number. Per-query fallback figures remain computable (never zeroed to read as free).
- Pricing staleness routes into the same path via `is_pricing_stale()`.
- Athena/Synapse serverless runs outside their verified region sets already resolve as flagged fallbacks, so the region guard flows through the new wiring with no `pricing.py` change. Their calculator docstrings no longer claim uniform flat-rate pricing.

Evidence:

- A Databricks serverless SQL config on an unoffered tier now yields `unavailable` with a populated warning (previously `normalized`, $4.40).
- Athena `sa-east-1` and Synapse serverless `brazilsouth` now yield `unavailable` instead of a flat $5.00/TB `normalized` cost; verified regions stay `normalized`.
- New tests in `tests/unit/core/cost/test_status_wiring.py` (7 cases, including a staleness probe and a one-warning-per-table check).

Deliberately unchanged:

- `benchbox/validation/bundle.py`: the hand-copied validator already enforces `unavailable` semantics and the contract shape is unchanged (validator parity passes).
- `EXPLORER_READ_MODEL_VERSION` stays at 10: no explorer-build column, view, or derivation changes, and `cost_status` is bundle-supplied data copied verbatim, so cached-snapshot invalidation follows bundle rebuilds as with every data correction.


joeharris76:
This branch is fully subsumed by merged work and now has zero diff vs develop: the fail-closed pick replayed merged #2195, and the region-guard picks were superseded by merged #2201 (regional tables plus the Sao Paulo/Brazil regional-rate and moon-region unlisted tests already on develop). All 334 cost tests pass on the reconciled tree. Closing as superseded; reopen if this branch is meant to carry follow-ups.

